PR #1, opened 2019-05-01
Requirements: Color/Emotion Picker

sjawhar:
## Purpose
Migrated requirements from GDrive to GitHub for team review and to kick-start development.

- [ ] Confirm number of colors
- [ ] Confirm list of emotions
- [ ] Is central display just a screen, or do we interact with it using some API?


docs/Color-Picker/Color-Picker.md
```diff
@@ -1 +1,24 @@
-TODO
\ No newline at end of file
+## Requirements
+
+### Part 1: App-Only
+* The audience will listen to the orchestra play four pieces of music.
+* Each audience member will use the app to pick from six available colors (RGB, CMY).
```
sjawhar: Need to follow-up with Dr. Steve Palmer in New Mexico for final color choices.

docs/Color-Picker/Color-Picker.md
```diff
@@ -1 +1,24 @@
...
+* It's possible that we will want audience members of the final rehearsal to be able to choose emotions instead.
+* The audience member can pick a new color at any time, but the app will give a gentle reminder every 20 seconds.
+* During the intermission (after the four songs), each audience member's app will display both their own choices as well as the overall audience distribution. Let's call this the "results" screen.
+* The app will move from one song to the next as the event progresses without user input
+
+### Part 2: App + Central Visualization
+* In the third section, the audience will listen to the orchestra play the Coriolan overture.
+* The audience will be randomly split into two groups.
+    * In group A, the functionality is the same as the app-only section.
+    * In group B, the audience member will instead be picking emotions: Joy, sadness, anger, fear, [disgust].
```
sjawhar: Not sure about disgust, might possibly need other secondary emotions. Need to finalize with Arseny.

How should we display the emotions? Text? Emojis?

docs/Color-Picker/Color-Picker.md
```diff
@@ -1 +1,24 @@
...
+* There will also be a central display in the event venue visualizing the audience's choices as they come in. This should function like a "mood ring" in the sense that it shows the most recent audience choices.
```
sjawhar: We should build our own, but there's a possibility that some super-awesome team might contribute one for the event. If that happens, ours can still be used in the app.

docs/Color-Picker/Color-Picker.md
cooperwalter: Psychology Today has a graphic showing a wheel of emotions. Not sure how accurate its structure is, but having all emotions in one place to think about might be useful. [What are Basic Emotions](https://www.psychologytoday.com/us/blog/hide-and-seek/201601/what-are-basic-emotions) 

Maybe a combination of emojis and text labels for them would allow quick identification of each option while maintaining clarity? For example: [Emoji Emotions](https://i.pinimg.com/originals/09/5a/ea/095aea060c73286d48b37a2c4dd63e56.jpg)

sjawhar: I like the idea of a grid of emotions, with emojis and text. If we have a color wheel, displaying emotions in a wheel as well might bias someone to choose the emotion that's in the same position in the wheel as the color that they chose on their first listen.

cooperwalter: I agree. I could see how that could present a bias issue.

docs/Color-Picker/Color-Picker.md
```diff
@@ -1 +1,285 @@
...
+]
+```
+
+#### Get aggregate choices
+Send `REQUEST_AGGREGATE_CHOICES`
+
+**Response**
+```js
+{
+  songs: [
```
sjawhar: We can manually log the start and end times of each song during the event. Then, during the intermission section where the user can see the aggregated results, this song list will be available.

docs/Color-Picker/Color-Picker.md
```diff
@@ -1 +1,285 @@
...
+}
+```
+
+#### User choices
+```js
+{
+  songId: String,           // Can be set to 'CONCERT_1' for the entire event
+  listenId: String,         // equal to userId
+  choices: {
+    [timestamp]: color,     // Key is the timestamp of the choice
```
sjawhar: Here's the thinking behind making this an object with timestamp as the key instead of an array: if there's a disconnection or a network hiccup leads to a repeated request, we won't have duplicate entries.

docs/Color-Picker/Color-Picker.md
```diff
@@ -1 +1,285 @@
...
+### During Event
+
+#### Song information
+```js
+{
+  songId: String,           // Partition key
+  listenId: 'SONG_INFO',    // Range key
+  title: String,            // Helpful for in-app displays
+  startTime: String,
+  endTime: String,
```
sjawhar: Like I mentioned above, we can populate these timestamp fields during the event.

docs/Color-Picker/Color-Picker.md
Borchardt: Is the central display an existing piece of equipment at the concert hall, or is it something that we're planning to set up ourselves? If the former, do we know whether there's some API we'll need to use to communicate with it? Or will it be as simple as plugging an input cable into some device to mirror app display?

docs/Color-Picker/Color-Picker.md
```diff
@@ -1 +1,285 @@
...
+
+### Part 3: After Event
+* After the event, the user should also be able to re-listen to each song and re-submit color or emotion choices.
+* The first time a user listens to a song, they should only have the ability to choose colors and should be blind to the existence of the emotion picker. The user can then choose which they'd like to do in future playbacks.
+* The user should be able to see their choices for each instance of listening to the song compared to the total audience. There are two parts to this:
+    * A static comparison of how the responses changed over time (the "results" screen)
+    * A dynamic comparison used while listening to a song (the live visualization)
+* The total audience distribution will at first contain only the concert submissions, but an admin should be able to update this to include additional data points.
+
+## API Contract
```
Borchardt: Not seeing anything in here for the "goosebumps" picker that we discussed. Are we still targeting that functionality? If so, would it be useful for any of the API's here to have bio-sensor deviceId fields?

docs/Color-Picker/Color-Picker.md
```diff
@@ -1 +1,285 @@
...
+```js
+{
+  token: String,    // access token from AWS Cognito
+}
+```
+
+#### Choose a color
+Send `COLOR_CHOSEN`
+```js
+{
```
Borchardt: Curious: why listenId in the after event color selection but no listenId here?

docs/Color-Picker/Color-Picker.md
Borchardt: :+1:

docs/Color-Picker/Color-Picker.md
Borchardt: Makes sense to me!

docs/Color-Picker/Color-Picker.md
sjawhar: This doc is just for the color picker. I wanted to take them one piece at a time.

docs/Color-Picker/Color-Picker.md
sjawhar: In the after-event functionality, a "listen" is an entity. You can start a new listen, review past listens, etc. During the event, it's all just one listen, so the listenId in the database is just the userId.

docs/Color-Picker/Color-Picker.md
sjawhar: That's a good question. I'm hoping it's the latter. Arseny mentioned "screens".

docs/Color-Picker/Color-Picker.md
```diff
@@ -1 +1,285 @@
...
+    {
+      id: String,
+      startTime: String,
+      endTime: Number,
+    },
+    ...
+  ],
+  choices: [
+    {
+      timestamp: Number,
```
Borchardt: Minor: is the Number type for timestamp here intentional, or should it be String to match others?

docs/Color-Picker/Color-Picker.md
```diff
@@ -1 +1,285 @@
...
+**Response**
+```js
+[
+  {
+    id: String,
+    mediaUrl: String,
+    title: String,
+    artist: String,
+    length: Number,
+  }
```
Borchardt:
```suggestion
  },
  ...
```

docs/Color-Picker/Color-Picker.md
```diff
@@ -1 +1,285 @@
...
+  choices: [
+    {
+      timestamp: Number,
+      colors: {
+        COLOR_RED: Number,
+        COLOR_BLUE: Number,
+        COLOR_GREEN: Number,
+        ...
+      },
+    },
```
Borchardt:
```suggestion
    },
    ...
```

docs/Color-Picker/Color-Picker.md
```diff
@@ -1 +1,285 @@
...
+```js
+{
+  songId: String,
+  listenId: 'AGGREGATE',
+  choices: {
+    [timeOffset]: {
+      COLOR_RED: Number,
+      COLOR_BLUE: Number,
+      COLOR_GREEN: Number,
+      ...
```
Borchardt: I'm guessing that since we may have emotions as well as colors at this point, the choices list here may contain keys for emotions in addition to keys for colors. Is that right?

docs/Color-Picker/Color-Picker.md
Borchardt: Maybe overthinking this, but if there will be text in the app (like for emoji labels here), do we need to care about supporting more languages than just English?

docs/Color-Picker/Color-Picker.md
sjawhar: Emotion was the next thing I was going to weave in, actually wanted to ask your opinion. Do you think there should be a `choiceType` field (CHOICE_COLOR, CHOICE_EMOTION) and have any fields I might have called "color" be called "choice"?

docs/Color-Picker/Color-Picker.md
sjawhar: Should be String, thanks! Started with time offset, then changed based on your suggestion from the email chain.

docs/Color-Picker/Color-Picker.md
Borchardt: Got it. Do we care about being able to distinguish between during-concert and after-concert selections? If so, would we have to do so by checking whether or not a given entry has a compound key? Or do the separate headings for during- and after-concert data mean that they'll be stored in different tables?

Borchardt: Similarly: am I correct in understanding that we plan to set up separate API's for during- and after- event functionality? Or do the different headings there just show what we expect data to look like for each use case?

docs/Color-Picker/Color-Picker.md
Borchardt: If color and emotion selection data will be sharing a table, my gut is that it might be best practice to have an attribute (e.g. choiceType like you've suggested above) that explicitly says which type of data any given entry is. Honestly, though, unless there's a big performance difference between querying based on attribute starts with vs. querying on attribute equals (choiceType == CHOICE_COLOR vs key starts with COLOR_), I don't know that it matters a ton. Fine leaving it whichever way you prefer.

Borchardt: (Separate, but maybe related: if you're working on emotion API contracts as well as data storage, I'd suggest that even though emotion and color selections will share a bunch of the same fields, it probably makes sense to stand up separate API's rather than extend the existing ones. Improves readability, makes it easier to keep track of necessary test coverage, etc)

docs/Color-Picker/Color-Picker.md
sjawhar: My plan was:
* Develop, test, and launch functionality for event
* Develop and test functionality for after event
* Have the event
* Launch after-event functionality and migrate data

docs/Color-Picker/Color-Picker.md
sjawhar: > If color and emotion selection data will be sharing a table, my gut is that it might be best practice to have an attribute (e.g. choiceType like you've suggested above) that explicitly says which type of data any given entry is. 

I agree, let's do that.

As for separate API's, I could be overlooking something, but I don't think it will be necessary. Seems like just a couple if statements to me, but I think we can call that one as we develop it.

sjawhar:
I've added the emotion functionality. The overarching theme is simplicity for the event (e.g. entire data for each user is in one record, just log the data and let the client handle some of the parsing) and explicitness for afterwards. If nobody sees any major problems, we can merge this PR and start another one for the architecture.

docs/Color-Picker/Color-Picker.md
Borchardt: Sure. No big deal either way, especially for a project this small

docs/Color-Emotion-Picker/Color-Emotion-Picker.md
```diff
@@ -218,9 +241,15 @@ Send `REQUEST_GET_CHOICES`
...
-    [timestamp]: color,     // Key is the timestamp of the choice
-    Number: String,         // Value is the color choice
+  colors: {
+    Number: String,         // Key is the timestamp of the choice
+    [timestamp]: color,     // Value is the color choice
+    ...
+  },
+  emotions: {
+    Number: String,         // Key is the timestamp of the choice
+    [timestamp]: color,     // Value is the color choice
```
Borchardt: Minor: value here is emotion, not color

docs/Color-Picker/Color-Picker.md
cooperwalter: Great point, Jackson. Quite possibly. My understanding is that French is the primary language in Lausanne. Something to check in with Arseny about.

docs/Color-Picker/Color-Picker.md
cooperwalter: To clarify: Audience members will only be able to pick a single color (or emotion) at a time, correct?

PR #3, opened 2019-05-18
Design: Mental Imagery, Chills, and Concert/After-Concert

sjawhar:
## Purpose
We've designed the functionality for the color/emotion picker, but this Friday's call made me realize the need to figure out the rest soon-ish. So I've taken a crack at it.

## Approach
My original design had everything happening over a websocket connection because of the realtime requirements of the concert. However, I realized that _only_ the concert needed realtime functionality. Everything afterwords would work just fine with REST APIs. So I separated the two, which gives us some nice benefits:
* Can compartmentalize the logic and deploy the two components separately
* Don't need a standing websocket server after the event, keeping costs down

I also fleshed out the functionality for the other apps, to make sure that we're not missing anything big. It seems like everything follows fairly straightforwardly from the work we've already done. Let me know if you think I missed anything.

- [ ] Finalize the requirements of mental imagery, chills, and composition apps


docs/Usage/Concert/Concert.md
```diff
@@ -0,0 +1,224 @@
...
+| Emotions          | Artist       | Song                                                 | Duration |
+|-------------------|--------------|------------------------------------------------------|----------|
+| Serenity/Calm     | Fauré        | Pavane - initial excerpt                             | 2:30     |
+| Sadness/Nostalgia | Grieg        | Elegiac Melodies - No. 2, Last Spring excerpt        | 2:00     |
+| Happiness/Joy     | Beethoven    | Symphony No. 7 - 4th movement - finale excerpt.      | 2:00     |
+| Anger             | Shostakovich | Chamber Symphony in C minor, Op. 110a - 2nd movement | 3:00     |
+
+* The audience will listen to the orchestra play the above four pieces of music.
+* Each audience member will use the app to pick from six available colors (RGB, CMY).
+* It's possible that we will want audience members of the final rehearsal to be able to choose emotions instead.
```
Borchardt: Not a huge deal, but let's make sure not to lose track of this. Will make our lives to know one way or the other sooner rather than later.

docs/Usage/Concert/Concert.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Websocket API
+### Report a choice
+Send `CHOICE_MADE`
+```js
+{
+  choiceType: String,           // CHOICE_COLOR || CHOICE_EMOTION || CHOICE_CHILLS || CHOICE_IMAGERY
+  choice: String || Boolean,    // Boolean for chills, String for the rest
+  timestamp: String,
+}
```
Borchardt: Thinking about displaying users' choices back to them during the intermission. Do we need some kind of userId (access token?) in here? Or will we store user choices in memory or get them some other way?

sjawhar: The mobile app can store the users' choices, I don't see a need for the back-end to repeat them back.

Also, the first thing the app does when opening the websocket connection is send a JWT.  At that point, the server can save user ID as a property on the connection object. So either way, we're covered.

Borchardt: Got it. Storing them in the app was what I meany by in memory. Works for me.

PR #7, opened 2019-05-29
Issue 4: User choices are written to Dynamo

### LINKED ISSUE #4: User choices are read from an SNS topic and written to a DynamoDB table
image

## Expected Behavior
When a user makes a choice during a concert, they send the `CHOICE_MADE` event. This gets pushed to an SNS topic and should be consumed by a lambda function to persist the user's choice to the table.

### Event Data
```js
{
  userId: String,
  choiceType: String,       // CHOICE_COLOR | CHOICE_EMOTION_HAPPINESS | CHOICE_EMOTION_AGITATION | CHOICE_CHILLS
  choice: String || Number, // String for color, Number for the rest
  timestamp: String,
}
```
### DB Record Structure
```js
{
  recordId: ['CHOICE', userId].join('$'),
  colors: {
    Number: String,         // Key is the timestamp of the choice
    [timestamp]: color,     // Value is the color choice
    ...
  },
  emotionType: String,
  emotions: {
    Number: String,         // Key is the timestamp of the choice
    [timestamp]: emotion,   // Value is the emotion choice
    ...
  },
  chills: {
    Number: Number,                // Key is the timestamp of the event
    [timestamp]: chillsIntensity,  // Value is numerical intensity of chills
    ...
  },
}
```

## Tasks
- [ ] Tests
- [ ] Code
- [ ] CloudFormation Resources

Borchardt:
## Purpose
This is a WIP pull request.
Closes #4 

## Approach
Updates handler in writer.py to process events (SNS notifications) and write to Dynamo according to contents

## Testing
Unit tests pass. Run the following command from within the lambdas directory.
```FUNCTION_NAME=$FUNCTION_NAME docker-compose -f docker-compose.util.yml run --rm pipenv run pytest choice/test/test_writer.py```

## Open Questions
* Is unit test coverage sufficient?
* General approach and code quality/style. This is the first reviewed code for this repository, so it will set the tone for future pull requests.

## Pre-Merge TODOs
* Instructions for testing locally with JSON test event and local DB (and confirmation of local test success)
* CloudFormation templates

### CODE REVIEW

File: lambdas/choice/src/writer.py [hunk 1]
```diff
@@ -1,6 +1,89 @@
...
+    # If choice type not recognized, return 422
+    if choice_type not in choice_types:
+        return {'statusCode': 422}
+    # Else get database key for choice type
+    choice_key = choice_type_to_choice_key[choice_type]
+
+    # If imagery choice, just add image word
+    if choice_key == choice_type_to_choice_key[CHOICE_IMAGERY]:
+        choice_value = choice
+        response = table.update_item(
```
sjawhar: If you wanted this to be a bit more DRY, you could try:
```python
def get_update_args(choice):
    args = {
        'Key': {
            'songId': CONCERT_SONG_ID,
            'listenId': choice['listenId'],
        },
        'ReturnValues': 'NONE',
        'ExpressionAttributeNames' = { '#choice_key': choice_keys[choice['type']] }
    }

    if choice['type'] == CHOICE_IMAGERY:
        args['UpdateExpression'] = 'SET #choice_key = :choice_value'
        args['ExpressionAttributeValues'] = { ':choice_value': choice['value'] }
        return args

   # Handle the more complicated case. Don't need else, have an early return
   pass

def save_choice(choice):
    args = get_update_args(choice)
    table.update_item(**args)
```

lambdas/choice/src/writer.py
```diff
@@ -1,6 +1,89 @@
...
+
+    # Get data from event
+    listen_id = event[EVENT_USER_ID_KEY]
+    choice_type = event[EVENT_CHOICE_TYPE_KEY]
+    choice = event[EVENT_CHOICE_KEY]
+    timestamp = event[EVENT_TIMESTAMP_KEY]
+
+    # If choice type not recognized, return 422
+    if choice_type not in choice_types:
+        return {'statusCode': 422}
```
sjawhar: I've only works with Lambda functions in Node, so please correct me if it's different in Python.

This function is being invoked by SNS, so statusCode is not meaningful. In fact, returning anything indicates to SNS that the function completed successfully. Instead, what you want to do is raise an error. This will cause a retry and, after 3 failures, redirect the message to a dead letter queue (if configured).

lambdas/choice/test/test_writer.py
```diff
@@ -1,5 +1,134 @@
-from src.writer import handler
+import boto3
+import os
+import pytest
+from src.writer import write_to_dynamo
```
sjawhar: It seems to me like this is testing implementation, not behavior. What we really care about is what happens when the handler is called, right? Are you doing this so you can inject the DynamoDB client? That shouldn't be necessary with stubbing.

lambdas/choice/src/writer.py
```diff
@@ -1,6 +1,89 @@
...
+EVENT_CHOICE_KEY = 'choice'
+EVENT_TIMESTAMP_KEY = 'timestamp'
+
+CHOICE_COLOR = 'CHOICE_COLOR'
+CHOICE_EMOTION = 'CHOICE_EMOTION'
+CHOICE_CHILLS = 'CHOICE_CHILLS'
+CHOICE_IMAGERY = 'CHOICE_IMAGERY'
+
+CONCERT_SONG_ID = os.environ['EVENT_NAME']
+DYNAMO_DB_TABLE_NAME = os.environ['DYNAMO_TABLE_NAME']
```
sjawhar: Finally, I know this is a really small nitpick and please don't hate me for bringing it up, buuuut....

It's DynamoDB, not Dynamo. `DYNAMODB_TABLE_NAME`

lambdas/choice/src/writer.py
```diff
@@ -1,6 +1,89 @@
...
+    aws_access_key_id=os.environ['AWS_ACCESS_KEY_ID']
+)
+table = dynamodb.Table(DYNAMO_DB_TABLE_NAME)
+
 
 def handler(event, context):
-  print(event)
-  print(os.environ)
-  return { 'statusCode': 204 }
+    return write_to_dynamo(dynamodb, event)
```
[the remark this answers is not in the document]
Borchardt: Yeah. As you guessed below, the intent here wasn't to encapsulate, but to allow for client injection; something along the lines of do_stuff() may have been a more accurate name. Regardless, you're right that logic should be better encapsulated; will fix.

lambdas/choice/src/writer.py
Borchardt: No hate! Will fix.

lambdas/choice/src/writer.py
```diff
@@ -1,6 +1,89 @@
...
+    # Else append timestamp: choice
+    else:
+        choice_value = {timestamp: choice}
+        # If attribute not present, add an empty list, then append to it
+        response = table.update_item(
+            Key={
+                'songId': CONCERT_SONG_ID,
+                'listenId': listen_id
+            },
+            UpdateExpression='SET {} = list_append(if_not_exists({}, :empty_list), :choice_value)'.format(choice_key, choice_key),
```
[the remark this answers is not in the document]
Borchardt: Yep, you're absolutely right. Dumb reading on my part: I saw ellipses in the design doc and my mind went to list of maps rather than just a single map. Will fix.

lambdas/choice/src/writer.py
```diff
@@ -1,6 +1,89 @@
...
+    timestamp = event[EVENT_TIMESTAMP_KEY]
+
+    # If choice type not recognized, return 422
+    if choice_type not in choice_types:
+        return {'statusCode': 422}
+    # Else get database key for choice type
+    choice_key = choice_type_to_choice_key[choice_type]
+
+    # If imagery choice, just add image word
+    if choice_key == choice_type_to_choice_key[CHOICE_IMAGERY]:
```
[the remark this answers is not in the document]
Borchardt: Yeah, it is, and you're right that it's easier to read that way. Changing.

lambdas/choice/src/writer.py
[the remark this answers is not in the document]
Borchardt: Two follow-up issues here:

1: As I understand it, Dynamo doesn't allow insertion into maps for attributes that do not yet exist. This means that if a row hasn't yet been created for a user, or if an attribute has not yet been created for a given choice type, the update for that choice for that user will fail. Can we expect user rows and attribute columns to be created before invocation of this Lambda? If we don't already have a plan for that, it might make sense to do a conditional put during authentication.

2: Dynamo seems not to like non-String object keys. With that in mind, is it alright if we store timestamps as Strings rather than as Numbers?

lambdas/choice/src/writer.py
```diff
@@ -1,6 +1,89 @@
...
+def write_to_dynamo(table, event):
+
+    # Get data from event
+    listen_id = event[EVENT_USER_ID_KEY]
+    choice_type = event[EVENT_CHOICE_TYPE_KEY]
+    choice = event[EVENT_CHOICE_KEY]
+    timestamp = event[EVENT_TIMESTAMP_KEY]
+
+    # If choice type not recognized, return 422
+    if choice_type not in choice_types:
```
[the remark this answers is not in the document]
Borchardt: Fixed.

lambdas/choice/src/writer.py
```diff
@@ -1,6 +1,89 @@
...
+
 
 def handler(event, context):
-  print(event)
-  print(os.environ)
-  return { 'statusCode': 204 }
+    return write_to_dynamo(dynamodb, event)
+
+
+def write_to_dynamo(table, event):
```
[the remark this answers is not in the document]
Borchardt: Yeah, this was just to allow injection of a stubbed client. Should be resolved after completion of discussion on that comment, so will leave as is for now.

lambdas/choice/src/writer.py
```diff
@@ -1,6 +1,89 @@
...
+
+    # If imagery choice, just add image word
+    if choice_key == choice_type_to_choice_key[CHOICE_IMAGERY]:
+        choice_value = choice
+        response = table.update_item(
+            Key={
+                'songId': CONCERT_SONG_ID,
+                'listenId': listen_id
+            },
+            UpdateExpression='set {} = :choice_value'.format(choice_key),
```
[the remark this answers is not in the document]
Borchardt: Fixed.

lambdas/choice/src/writer.py
Borchardt: Nope, you're right. Fixed.

lambdas/choice/test/test_writer.py
Borchardt: Yeah, the import of the write_to_dynamo method here was to allow injection of a stubbed Dynamo client. I understand your concern about testing implementation vs. behavior, but can you point me toward whichever way you prefer to handle stub injection?

Borchardt:
Focused on src concerns for this revision, as there's ongoing conversation about approach to stubbing in tests.

For this revision, I was able to invoke the Lambda locally successfully with the command ```docker-compose -f docker-compose.development.yml run --rm invoke-lambda lambdas/choice/test/writer_event.json ChoiceWriterFunction```. Querying my local DB after a couple of invocations produced the following output:

```
GetItem succeeded:

{'songId': 'CONCERT_LAUSANNE_2019', 'colors': {'99999991': 'chartreuse', '9999999': 'chartreuse'}, 'listenId': 'test user'}
```

However, I was only able to get this to work with environment variable values hardcoded. My env.local.json file looks as follows:

```
{
  "ChoiceWriterFunction": {
    "CHOICE_TABLE_NAME": "local-choices",
    "DYNAMODB_ENDPOINT": "http://db:8000",
    "EVENT_NAME": "CONCERT_LAUSANNE_2019",
    "DYNAMODB_TABLE_NAME": "local-colors"
  }
```

Anything else that I should need to update to get those values picked up?

lambdas/choice/src/writer.py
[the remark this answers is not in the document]
sjawhar: > Can we expect user rows and attribute columns to be created before invocation of this Lambda?
I was planning to have the record for the user created on connect, with `colors`, `emotions`, etc. initialized as empty maps.

> is it alright if we store timestamps as Strings rather than as Numbers?
Good catch! I don't see a problem with that.

lambdas/choice/src/writer.py
```diff
@@ -1,6 +1,78 @@
...
+CONCERT_SONG_ID = os.environ['EVENT_NAME']
+DYNAMODB_TABLE_NAME = os.environ['DYNAMODB_TABLE_NAME']
+
+choice_type_to_choice_key = {
+    CHOICE_COLOR: 'colors',
+    CHOICE_EMOTION: 'emotions',
+    CHOICE_CHILLS: 'chills',
+    CHOICE_IMAGERY: 'imagery'
+}
+choice_types = choice_type_to_choice_key.keys()
```
sjawhar: `choice_types` is not being used. I'm going to add a linter and auto-formatter so we don't have to think about these things.

lambdas/choice/test/test_writer.py
sjawhar: I might be making assumptions re: stubbing based on my experience with Node, but I _think_ you shouldn't need to inject the stubs:
```python
from src.writer import handler, dynamodb
from botocore.stub import Stubber

def test_function():
    with Stubber(dynamodb.meta.client) as stub:
        # Test things!
        pass
```

sjawhar:
For your environment variable question: make sure all the environment variables are declared on the function (or in the globals section) in the cloudformation file. SAM reads that to know which variables to look for in env.local.json

lambdas/choice/src/writer.py
```diff
@@ -1,6 +1,78 @@
...
+
+dynamodb = boto3.resource(
+    'dynamodb',
+    region_name=os.environ['AWS_DYNAMO_REGION'],
+    endpoint_url=os.environ['AWS_DYNAMO_ENDPOINT'],
```
sjawhar: Outside of local development, we're going to want to omit `endpoint_url`. Might want to do a `dynamodb_args` dict, add `endpoint_url` is `os.environ[DYNAMODB_ENDPOINT]` is set, then `boto3.resource('dynamodb', **dynamodb_args)`

lambdas/choice/src/writer.py
```diff
@@ -1,6 +1,78 @@
...
+    aws_secret_access_key=os.environ['AWS_SECRET_ACCESS_KEY'],
+    aws_access_key_id=os.environ['AWS_ACCESS_KEY_ID']
```
sjawhar: These credentials are only needed in local development, right?

lambdas/choice/test/test_writer.py
Borchardt: Nope, you're totally right. Importing the dynamodb variable was the solution. Thanks for the tip!

lambdas/choice/src/writer.py
Borchardt: Good suggestion. Fixed.

lambdas/choice/src/writer.py
Borchardt: Actually, now that we have the DynamoDB client stubbed properly, I don't think they're needed at all. Removed all except for endpoint, which has been updated per suggestion above.

Borchardt: Actually, I think that we can get rid of the two access key parameters entirely. I've updated the other two according to what you suggested for DYNAMODB_ENDPOINT above.

lambdas/_local-setup/setup-db.js
```diff
@@ -9,7 +9,7 @@ const main = async () => {
   try {
     await dynamoDb
       .createTable({
-        TableName: 'local-colors',
+        TableName: 'local-choice
```
sjawhar:
```suggestion
        TableName: 'local-choice',
```

cloudformation.yml
```diff
@@ -85,7 +85,9 @@ Resources:
             Topic: !Ref ChoiceMadeSnsTopic
       Environment:
         Variables:
-          CHOICE_TABLE_NAME: !Ref ChoiceTable
+          DYNAMODB_TABLE_NAME: !Ref ChoiceTable
+          DYNAMODB_ENDPOINT:
```
sjawhar: This is set under `Globals`, shouldn't be needed here.

Borchardt: Good catch. Fixed.

lambdas/choice/test/color_chosen_event.json
```diff
@@ -0,0 +1,6 @@
+{
+  "userId": "user id",
```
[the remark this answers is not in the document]
Borchardt: Good callout. I hadn't realized that SNS events come through with message JSON not at the top level. Fixed.

lambdas/choice/src/writer.py
```diff
@@ -1,7 +1,71 @@
...
+    CHOICE_COLOR: "colors",
+    CHOICE_EMOTION: "emotions",
+    CHOICE_CHILLS: "chills",
+    CHOICE_IMAGERY: "imagery",
+}
+
+dynamodb_args = {}
+if os.environ.get("DYNAMODB_ENDPOINT") is not None:
+    dynamodb_args["endpoint_url"] = os.environ["DYNAMODB_ENDPOINT"]
+if os.environ.get("AWS_DYNAMO_REGION") is not None:
```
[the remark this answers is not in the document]
Borchardt: Yep, sure can. Fixed.

cloudformation.yml
```diff
@@ -85,7 +85,9 @@ Resources:
...
+          EVENT_NAME:
```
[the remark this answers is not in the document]
Borchardt: Done.

Borchardt:
Our linter doesn't like the string escaping in the event messages in test_writer.py. I think the correct way to resolve that is to move those test messages into separate JSON files, but wanted to get your take before doing so.

sjawhar:
> Our linter doesn't like the string escaping in the event messages in test_writer.py. I think the correct way to resolve that is to move those test messages into separate JSON files, but wanted to get your take before doing so.

Yeah, you could use fixtures. Or you could use a helper `get_event()` function that takes a dictionary and returns the whole wrapped event `SNS.Records[0].Message.HowFarDown.DoesThisThingGo = json.dumps(message)`

PR #26, opened 2019-07-11
Chore/lint and build

sjawhar:
## Purpose
Linting, auto-formatting, and build variables

## Approach
* `eslint-config-universe` for setting up eslint and prettier
* `react-native-dotenv` for passing build variables
* Reorganized all client code under client/src
* Added client-lint build process

## Testing
@cooperwalter @Borchardt Please make sure you can still build and run the app!


cooperwalter:
Looks good to me. The app still builds just fine. 

I also went ahead and fixed the two errors that were being caught by the linter. One was a duplicate style declaration; the other was due to WebSocket being a global variable (and thus eslint caught it as undefined). So I added WebSocket as a global variable to the eslint config.

Approving.

cooperwalter:
Hmm, bizarre. The error is gone. @Borchardt does the app work fine on your end?

Borchardt:
I'm able to build and run the app without issue as well. Looks good to me!

PR #27, opened 2019-07-16
todo/app navigates to correct screen

### LINKED ISSUE #22: On the CONNECTED and EVENT_STAGE_CHANGED websocket events, the app navigates to the correct screen
## Expected Behavior
After connecting, the websocket API will send the following event.

```js
{
  event: 'CONNECTED',
  data: {
    choiceType: String, // CHOICE_COLOR | CHOICE_EMOTION_HAPPINESS | CHOICE_EMOTION_AGITATION
    choiceInverted: Boolean,
    stageId: String,
    ... // other stage-specific attributes
  },
}
```

This event contains the current stage of the event. Based on the value of `stageId`, the app should then navigate to the correct screen (if it is not there already):
* STAGE_CHOICE_IMAGERY: show the mental imagery form
* STAGE_CHOICE_COLOR_EMOTION: show the color/emotion picker
* STAGE_CHOICE_CHILLS: show the chills UI
* STAGE_END: show the results screen

`choiceType` and `choiceInverted` should be stored for later (see #25).

The app should also respond to the EVENT_STAGE_CHANGED event (same body, just different `event` value and no `choiceType` or `choiceInverted`) in the same way.

## Context
We need to be careful to do this in a way such that the user's experience is not disrupted (or even worse, data loss) if the app disconnects and reconnects during the concert.

None of these screens need to be finished as part of this story. They can be empty stubs. But the app should navigate between them.

cooperwalter:
## Purpose
Allow app to navigate to the correct screen upon receiving CONNECTED and EVENT_STAGE_CHANGED messages. Screens that don't exist yet have been stubbed. Navigation is handled by a navigation service.

Closes #22.

## Approach
Redux and redux-persist are used to store choiceType and choiceInverted, as well as the last known screen the user was on (in case the app quits unexpectedly). Event listeners are placed in App.js to respond to events emitted from the concert api, storing state in redux and navigating to the correct screen by use of the navigation service.


client/app/src/api/concertApi.js
```diff
@@ -36,9 +36,11 @@ const connect = idToken => {
     console.log('MESSAGE', message.data);
     try {
       const { event, data } = JSON.parse(message.data);
-      if (event === 'EVENT_STAGE_CHANGED') {
-        console.log('emitting SHOW_COLOR_PICKER event');
-        events.emit('SHOW_COLOR_PICKER');
+      if (event === "CONNECTED") {
```
sjawhar: If we're just going to mirror the event and data, why bother with the if statements?

client/app/src/state/reducer.js
```diff
@@ -0,0 +1,27 @@
+import actionTypes from "./actionTypes";
+
+const INITIAL_STATE = {
+  lastKnownScreen: "Welcome",
```
sjawhar: Is this going to be used for anything?

client/app/src/App.js
```diff
@@ -1,14 +1,76 @@
...
+};
+
+concertApi.on('WEBSOCKET_CONNECTED', data => {
+  console.log('stream in ws connected is:', data);
+  try {
+    const { stageId, choiceType, choiceInverted } = data;
+    console.log('choiceType in WS connected is:', choiceType);
+    store.dispatch(actions.setChoiceType(choiceType));
+    store.dispatch(actions.setChoiceInverted(choiceInverted));
+    handleStageNavigation(stageId);
```
sjawhar: This seems to be discarding all the stage-specific data and only preserving stageId. Do you have a plan for how to persist that additional information?

client/app/src/state/reducer.js
cooperwalter: I don't think it will be used for anything, given that the current stageId is provided on connection to the websocket. I'll remove it.

client/app/src/App.js
cooperwalter: You're right, it currently is. Will the data for one stage need to be stored after the concert progresses to the next stage? 

If it will need to be kept, then the stage data could be stored as a stageAttributes property of the root redux state, with keys that correspond to each stageId. E.g., to access the stage-specific data for STAGE_CHOICE_CHILLS would be accessed by state.stageAttributes.STAGE_CHOICE_CHILLS. And the attributes for the current stage could be accessed by state.stageAttributes[state.stageId] (see my responses to your comment above).

If the data can be discarded after progression to the next stage, then the stage-specific data could be stored by in state.stageAttributes that is overwritten on progression to the next stage. 

What do you think?

client/app/src/state/reducer.js
cooperwalter: However, it may make sense to store the current stage id, depending on whether the stage-specific attributes of previous stages will need to be kept for some reason (see below).

client/app/src/api/concertApi.js
cooperwalter: I had the idea in my head that naming the events emitted by the EventEmitter differently from the events emitted by the websocket would lead to less confusion and make debugging easier. But thinking about it now, I don't see how it's any better than mirroring the event and data. I think I'll go ahead and remove the if statements.

client/app/src/App.js
sjawhar: At the end of the app-guided part of the concert (during the intermission), the user should be presented with the "results screen" which shows their choices compared to the aggregate. The aggregate choices will be provided by the back-end, but the user's choices need to be persisted in the app to be displayed in this section. Where this information is stored, whether it's in a "stage-specific" part of the data store, is up to you. Does the app need to remember that the last stage was frequency 20 and timeout 5? No.

client/app/src/App.js
```diff
@@ -1,14 +1,75 @@
...
+    if (choiceType === 'CHOICE_COLOR') {
+      screen = 'Colors';
+    } else {
+      screen = 'Emotions';
+    }
+  } else if (stageId === 'STAGE_CHOICE_CHILLS') {
+    screen = 'Chills';
+  } else if (stageId === 'STAGE_END') {
+    screen = 'Results';
+  } else {
```
sjawhar: STAGE_WAITING is also a possibility: https://github.com/sjawhar/enchanted-brain/wiki/Concert#1---welcome

client/app/src/App.js
```diff
@@ -1,14 +1,75 @@
...
+import concertApi from './api/concertApi';
+import NavigationService from './navigation/NavigationService';
 import AppNavigator from './navigation/AppNavigator';
 import layout from './constants/Layout';
 import config from './config';
 
+// ** Event listeners ** //
+const handleStageNavigation = stageId => {
+  console.log('in handleStageNavigation');
+  const currentState = store.getState();
```
sjawhar: `currentState` isn't being used.
```suggestion
  const { choiceType } = store.getState();
```

sjawhar:
image
Just reporting a successful connection through the app

sjawhar:
Moving to #29

PR #28, opened 2019-07-16
add tests for callback function

### LINKED ISSUE #15: A lambda function POSTs callbacks to the API gateway
image

## Expected Behavior
A lambda function should consume from the callback SQS queues (see #12) to POST messages back to connected clients. `connectionId` can be obtained from the SQS queue name. The entire message payload should be POSTed.

## Possible Solution
~https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-how-to-call-websocket-api-connections.html~
[Much simpler](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigatewaymanagementapi.html#ApiGatewayManagementApi.Client.post_to_connection)

## Tasks
- [ ] Tests
- [ ] Code

Borchardt:
## Purpose
Add tests for CallbackFunction handler implemented during hackathon.
Closes #15

## Approach
Add test cases for successful invocation, invocation with invalid message, and invocation with client error

## Testing
`docker-compose -f docker-compose.util.yml run --rm pipenv run pytest`

- [X] Have you written tests for these changes? These changes are tests


concert/app/test/test_callback.py
```diff
@@ -0,0 +1,49 @@
...
+def test_callback():
+    event = get_event()
+
+    expected_params = {"Data": json.dumps(event_body), "ConnectionId": "123456="}
+
+    with Stubber(apigateway) as stub:
+        stub.add_response(
+            "post_to_connection", post_to_connection_success_response, expected_params
+        )
+        handler(event, None)
```
sjawhar: It seems like this test has no assertions. Shouldn't we at least be checking that the expected call was made? I think we can do [`stub.assert_no_pending_responses()`](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/stubber.html#botocore.stub.Stubber.assert_no_pending_responses). Otherwise this test could pass without any call being made to SNS at all.

concert/app/test/__init__.py
```diff
@@ -4,3 +4,4 @@
 os.environ["AWS_DEFAULT_REGION"] = "us-east-1"
 os.environ["EVENT_NAME"] = "CONCERT_LAUSANNE_2019"
 os.environ["DYNAMODB_TABLE_NAME"] = "testing-enchanted-brain"
+os.environ["API_URL"] = "https://abc.execute-api.us-east-1.amazonaws.com/Prod"
```
sjawhar: Take a look at my PR to see how you can setup test-specific initialization by using directories.

Borchardt: Done

concert/app/test/test_callback.py
Borchardt: Good point, and good suggestion. Done.

concert/app/test/__init__.py
```diff
@@ -1,6 +0,0 @@
-import os
-
-# setup dummy environment variables
-os.environ["AWS_DEFAULT_REGION"] = "us-east-1"
```
sjawhar: This one can stay here, it's not specific to any one function and is needed to keep boto3 from being a little bitch

PR #29, opened 2019-07-18
Embed mental imagery survey.

### LINKED ISSUE #6: Mental imagery survey is embedded in the mobile app
## Expected Behavior
When the app receives the `EVENT_STAGE_CHANGED` event and the new event stage is `MENTAL_IMAGERY`, the app should display the mental imagery survey form.

When the form is submitted, the app should return to the "thank you/waiting" screen.

## Tasks
- [ ] Event listener
- [ ] Embed form
- [ ] Waiting screen

### LINKED ISSUE #22: On the CONNECTED and EVENT_STAGE_CHANGED websocket events, the app navigates to the correct screen

cooperwalter:
## Purpose
Closes #6
Closes #22

## Approach
Qualtrics has an embedded data feature that allows hidden inputs to be populated by a query string appended to the survey URL. (See [this article](https://www.qualtrics.com/support/survey-platform/survey-module/survey-flow/standard-elements/embedded-data/).) Opening this URL in a web view allows the participant to complete the survey and for us to capture the user's ID.

Answering questions on Qualtrics doesn't change the URL. So, to listen for form submission, I configured Qualtrics to redirect the user to https://www.google.com. The WebView listens for navigation changes, and if the URL matches https://www.google.com, the app navigates the user to the thank you/waiting screen. Any URL should work, as long as it matches the URL the WebView is listening for.

**UPDATE** Using NeuroTech Foundation website instead of google

## Testing
After the app connects and receives EVENT_CONNECTED, send the following event through SNS:
```js
{
    "event": "EVENT_STAGE_CHANGED",
    "data": {
        "stageId": "STAGE_CHOICE_IMAGERY",
        "formUrl": "https://ucsf.co1.qualtrics.com/jfe/form/SV_eEBoIQ1RAp6UxdH"
    }
}
```

## Pre-Merge TODOs

- [x] ~Activate the Qualtrics survey and change the preview survey URL in the app to the final, published survey URL.~ Get formUrl from stage data
- [x] Squash that damn unhandled promise rejection from Amplify


sjawhar:
I follow the changes, but it looks like we need to finish #27 first before we can be sure they will be functional. Right?

cooperwalter:
Right! #27 should be finished before this PR.

Also, where will the uid for the user originate? Once #27 is done, I can add the uid to the redux state and set up the mental imagery web view to post it to the Qualtrics survey.

sjawhar:
You get the user ID from the Cognito ID token:
https://github.com/sjawhar/enchanted-brain/blob/566806e3958ebb768573fd30285088410de59240/client/app/src/screens/Welcome.js#L12-L14

cooperwalter:
If we already have the user's JWT persisted in redux state, should we skip calling Auth.currentSession() and instead immediately call concertApi.connect() with the stored JWT?

sjawhar:
Where you store the ID token is up to you. But remember that ID tokens expire (every hour, I think), so you'll need to have a mechanism for checking that the ID token in the store isn't expired and refresh it if it is.

client/app/src/screens/Results.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from 'react';
+import { Text, View } from 'react-native';
+import EStyleSheet from 'react-native-extended-stylesheet';
+
+class ResultsScreen extends Component {
+  render() {
+    return (
+      <View style={styles.container}>
+        <Text>I'll be the Results screen someday.</Text>
```
Borchardt: we're all cheering for you, little guy

Borchardt:
Approving. Will leave it up to you whether to iterate here for unhandled promise error or track that work in another PR.

sjawhar:
Fixed that pesky rejection. That was really bugging me.

PR #30, opened 2019-07-20
Authorizer Function

### LINKED ISSUE #16: A lambda function authorizes API Gateway requests
image

## Expected Behavior
Websocket requests must include a Cognito JWT in the `Authorization` header. A lambda function should:
* Verify the token in `event["headers"]["Authorization"]` is a valid JWT belonging to the correct Cognito User Pool.
* Verify that the user's groups `token["cognito:groups"]` includes the event group
* Create an IAM policy authorizing access to the necessary methods:
  * All users should be able to call $connect and $disconnect
  * Users in the Admin group should be able to call EVENT_STAGE_CHANGED
  * Users not in the Visualization or Admin groups should be able to call CHOICE_MADE
* Sets the user's username `token["sub"]` as the principalId of the policy
* Sets the following additional context variables:
  * `is${group} = True` where `group` is the name of a user group (e.g. `isAdmin`)

## Possible Solution
* See this page for python JWT libraries: https://jwt.io/
* [Verifiying a JSON Web Token](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-verifying-a-jwt.html)
* [Create a Lambda REQUEST Authorizer Function](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api-lambda-auth.html)

## Tasks
- [ ] Tests
- [ ] Code

sjawhar:
## Purpose
Closes #16

## Approach
- Uses the [python-jose](https://github.com/mpdavis/python-jose) package to verify JWTs
- Added visualization user pool group

- [x] Add python-jose to authorizer deployment package
- [x] Deploy to and test in development
- [x] Add `is${Group}` attributes to request context
- [ ] There's no checking that this person is actually authorized to participate in the concert (i.e. is a ticket holder currently sitting in the concert hall). I think the easiest way to do that is to provide a code at the start of the concert (on the big screen) which people enter into the app to connect, and this code should be sent as an additional header in the request. Thoughts?


concert/app/src/functions/authorizer/index.py
```diff
@@ -1,36 +1,96 @@
...
+response = urlopen(
+    "https://cognito-idp.{}.amazonaws.com/{}/.well-known/jwks.json".format(
+        COGNITO_REGION, COGNITO_USER_POOL_ID
+    )
+)
+JWKS_KEYS = json.loads(response.read())["keys"]
 
+CHOICE_BUCKETS = [
+    "CHOICE_COLOR",
+    "CHOICE_COLOR",
```
Borchardt: Duplication of line above. Should this be CHOICE_CHILLS?

concert/app/src/functions/authorizer/index.py
```diff
@@ -1,36 +1,96 @@
...
+        if kid == JWKS_KEYS[i]["kid"]:
+            key_index = i
+            break
+    if key_index == -1:
+        return None
+
+    return jwt.decode(token, JWKS_KEYS[key_index], audience=COGNITO_APP_CLIENT_ID)
+
+
+def get_user_context(user_id, groups):
```
Borchardt: It's not obvious to me why we're doing the modular arithmetic below in the way that we're doing it. Would you mind explaining?

concert/app/src/functions/authorizer/index.py
sjawhar: No. 50% chance color, 25% chance happiness, 25% agitation

Borchardt: Ah. Got it.

concert/app/src/functions/authorizer/index.py
sjawhar: We need two random variables: choiceType and choiceInverted. 50% of people should be color, 25% happiness, 25% agitation. That's the first modulo. We also want 50% split for inverted/not. That's the second modulo. So there are eight groups: four are color, two are happiness (inverted/not), two are agitation (inverted/not).

Borchardt:
Looks clean.

Re: audience authorization: I'd love to avoid making audience members enter a code if possible, but I haven't yet been able to come up with a better alternative. I'll keep thinking on it, but think that's a reasonable plan going forward.

concert/app/src/functions/authorizer/index.py
Borchardt: Roger. Thanks for the explanation.

PR #32, opened 2019-07-23
OnDisconnect Function

### LINKED ISSUE #14: When a user disconnects from the gateway, their callback SQS queue is deleted
image

## Expected Behavior
Listening on the $disconnect route, a lambda function should delete the SQS queue created for the connection (see #12). If this was implemented using a CloudFormation stack, this is as simple as deleting the stack for that connection.

If this was not implemented using a CloudFormation stack, then the function needs to:
* Delete the Lambda event source mapping
* Delete the SNS subscription(s) (two if it's a visualization queue)
* Delete the SQS queue

## Tasks
- [ ] Tests
- [ ] Code

sjawhar:
## Purpose
Closes #14 

## Approach
Like #31, direct resource management. Lambda event source mappings take some time to recognize they're no longer needed, so I put in an exponential backoff retry.

### DIFF
--- a/concert/app/src/functions/on_disconnect/index.py
+++ b/concert/app/src/functions/on_disconnect/index.py
@@ -16,8 +16,7 @@ client_lambda = boto3.client("lambda")
 client_sns = boto3.client("sns")
 client_sqs = boto3.client("sqs")
 
-dynamodb = boto3.resource("dynamodb")
-table = dynamodb.Table(DYNAMODB_TABLE_NAME)
+table = boto3.resource("dynamodb").Table(DYNAMODB_TABLE_NAME)
 
 
 def handler(event, context):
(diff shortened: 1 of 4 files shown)


concert/app/src/functions/on_disconnect/index.py
```diff
@@ -29,7 +28,7 @@ def handler(event, context):
 
     resources = table.get_item(Key=connection_key)["Item"]
 
-    for subscription_arn in resources[ATTR_CONNECTION_SNS_SUBSCRIPTION_ARNS]:
+    for subscription_arn in sorted(resources[ATTR_CONNECTION_SNS_SUBSCRIPTION_ARNS]):
```
Borchardt: Just out of curiosity, why sorted here?

sjawhar: Because the test stubber is annoying and expects the calls in a specific order :(

Issue #17, opened 2019-06-26
When a user makes a choice, that choice is published to choice-made SNS, Kinesis, and callback SNS


Borchardt:
![image](https://user-images.githubusercontent.com/5074378/60968199-cb7eb000-a31c-11e9-925b-9a0b1fd350c3.png)

## Expected Behavior
When a user makes a choice through API Gateway, that choice should be processed by a Lambda which does three things:

1: Publish the choice to the ```choice-made``` SNS topic
2: Push the choice to a Kinesis Firehose Stream
3: Publish a message to the `callback-visualization` SNS topic

## Websocket Message Format
```
{
  event: 'CHOICE_MADE',
  data: {
    choiceType: String,
    choice: String || Number,
    timestamp: String
  }
}
```

## SNS Message Format
Data from websocket message, plus the `userId` from `event["requestContext"]["authorizer"]["principalId"]`.

## Context
When a user choice is made, it needs to be routed so that it can be persisted, processed, and added to the visualization. This Lambda handles that routing. For more information on the components it feeds to, see #4, #5, and #12.

## Tasks
- [ ] Tests
- [ ] Code
- [ ] CloudFormation Resources

PR #33, opened 2019-07-24
Issue5

### LINKED ISSUE #5: A Kinesis Data Analytics function aggregates user choices
image

## Expected Behavior
When a user makes a choice during a concert, they send the CHOICE_MADE event. This gets pushed to a Kinesis Firehose stream. In addition to persisting all data to an S3 bucket, this stream also feeds a Kinesis Data Analytics application that aggregates the choices.

### Incoming Record
```js
{
  userId: String,
  choiceType: String,          // CHOICE_COLOR || CHOICE_EMOTION || CHOICE_CHILLS
  choice: String || Number,    // String for color, Number for the rest
  timestamp: Number,
}
```

### Destination Stream Columns
```js
{
  CHOICE_TIME: Number,
  CHOICE_TYPE: String,
  CHOICE_SUM: Number,
  CHOICE_COUNT: Number,
}
```
## Context
Aggregated data will be used to provide a visualization of user choices in the app during the intermission.

## Possible Solution
Need to keep in mind that there will be multiple songs during the event. There should be some way of distinguishing which song the choices belong to. Maybe by also keep track of event stage changes in another stream and joining on that?

## Tasks
- [ ] Kinesis data analytics SQL

Borchardt:
## Purpose
Partially closes #5 

## Approach
This commit does a few things:
* Add CloudFormation resources for a Kinesis Data Analytics Application, a Firehose Delivery Stream, an S3 bucket destination (for the delivery stream), A CloudWatch LogGroup and LogStream for the delivery stream, and roles/policies to grant all necessary permissions
* Add a SQL query (in the ApplicationCode property of the Data Analytics Application) which tracks running sums and choice counts for each numerical choice type grouped by choice timestamp over the past minute

## Screenshots
image
image
(Note: Counts for both choice type reset after the fourth row. This is because more than a minute elapsed between input between the fourth and fifth rows. Also note that CHOICE_TIME timestamp values are all the same simply because that's the way they were input.)

## Testing
End to end testing carried out by deploying the updated CloudFormation template to my own AWS account, connecting to the WS server, sending choice data, and watching results appear in Kinesis Data Analytics (see screenshots above). New unit tests also pass.

- [ ] How much of what we desired from this issue is achieved by this work? Color choice data is ignored completely here (because of uncertainty over desired behavior), so follow-up work will definitely need to be done there. Are any other follow-ups necessary?
- [ ] Do the docs need to be updated? Yes, depending on answers to questions above


concert/cloudformation.yml
```diff
@@ -383,6 +386,117 @@ Resources:
...
+        - PolicyName: firehose_delivery_policy
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action: s3:*
+                Resource: '*'
+              - Effect: Allow
+                Action: logs:PutLogEvents
+                Resource: !Sub arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:/${Environment}-enchanted-brain-aggregateChoiceDataStreamLogs:*
```
sjawhar: You can use `!GetAtt AggregateChoiceDataStreamLogGroup.Arn` to simplify this and avoid typos

concert/cloudformation.yml
```diff
@@ -383,6 +386,117 @@ Resources:
...
+              Service: firehose.amazonaws.com
+            Action: sts:AssumeRole
+      Policies:
+        - PolicyName: firehose_delivery_policy
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action: s3:*
+                Resource: '*'
```
sjawhar: :open_mouth: 

No god mode, please

concert/cloudformation.yml
```diff
@@ -383,6 +386,117 @@ Resources:
...
+
+  AggregateChoiceDataStreamLogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: !Sub ${Environment}-enchanted-brain-aggregateChoiceDataStreamLogs
+      RetentionInDays: 30
+  AggregateChoiceDataStreamLogStream:
+    Type: AWS::Logs::LogStream
+    Properties:
+      LogGroupName: !Sub ${Environment}-enchanted-brain-aggregateChoiceDataStreamLogs
```
sjawhar: You can use `!Ref AggregateChoiceDataStreamLogGroup` to simplify this and avoid typos

concert/cloudformation.yml
```diff
@@ -383,6 +386,117 @@ Resources:
...
+        BucketARN: !GetAtt AggregateChoiceDataS3Bucket.Arn
+        BufferingHints:
+          IntervalInSeconds: 60
+          SizeInMBs: 50
+        CompressionFormat: UNCOMPRESSED
+        RoleARN: !GetAtt AggregateChoiceDataStreamRole.Arn
+        CloudWatchLoggingOptions:
+          Enabled: true
+          LogGroupName: !Sub ${Environment}-enchanted-brain-aggregateChoiceDataStreamLogs
+          LogStreamName: !Sub ${Environment}-enchanted-brain-aggregateChoiceDataStreamLogs
```
sjawhar: You can use !Ref to simplify both these options and avoid typos.

concert/cloudformation.yml
```diff
@@ -302,9 +302,12 @@ Resources:
         Variables:
           CALLBACK_VISUALIZATION_SNS_TOPIC_ARN: !Ref CallbackVisualizationSnsTopic
           CHOICE_MADE_SNS_TOPIC_ARN: !Ref ChoiceMadeSnsTopic
+          AGGREGATE_CHOICE_DELIVERY_STREAM_NAME: !Sub ${Environment}-AggregateChoiceDataStream
       Policies:
         - SNSPublishMessagePolicy:
             TopicName: !GetAtt ChoiceMadeSnsTopic.TopicName
+        - FirehoseWritePolicy:
+            DeliveryStreamName: !Sub ${Environment}-AggregateChoiceDataStream
```
sjawhar: You can use !Ref to simplify this, indicate a dependency, and avoid typos.

Borchardt:
For the query, the intent (and I believe the behavior) here is to aggregate not just across the last minute, but to aggregate across the last minute grouping by both choice type and choice timestamp. Unfortunately, I don't think it's possible to get rid of the temporal bound on the aggregation. As I understand it, Kinesis Data Analytics doesn't allow you to use `GROUP BY` (or `PARTITION BY`) clauses without adding monotonic bounds (typically the ROWTIME attribute specifying when your data entered the source stream) on the number of rows to be processed. The one minute preceding interval here was chosen as a fairly conservative bound guess on choice timestamps: after a minute, we can be fairly confident that we won't be seeing any more entries for a timestamp we've seen in the past (at least for actual use data), so the authoritative entry for any choice type/choice timestamp combination should just be the last one for it in the stream.

Having said all that, I think that we may be able to get better results by switching to a different aggregation window type. Kinesis Data Analytics allows aggregation over three different time window types: tumbling, sliding, and staggered. Tumbling means that every <whatever interval of time> after the application starts, results are computed across the last  <whatever interval of time>. This seems undesirable, because choices occurring near window boundaries may have be split across two windows. Sliding, which is what we're using here, uses rolling windows rather than fixed ones; this is why we're seeing multiple entries (with monotonically increasing counts) for each choice type/choice timestamp combination.  Instead of using a fixed or rolling interval, the third option type, staggered, maintains separate windows of specified temporal length for each unique partition key beginning upon receipt of new partition key values. I think that switching to a staggered window here may make sense, as it would keep us within the same accuracy threshold (entries for the same timestamp spanning more than a minute in real time from the stream's perspective will break our calculations) , but will remove the duplicate entries for each choice type/choice timestamp combination within each window.

With all that said, I'm more than happy to iterate on this with you (in a hackathon setting or otherwise). If any of the above changes your stance on accuracy of the query, it will also be pretty easy for me to make updates to get chills and color choices working (within the paradigm I described above) as you described; the only reason they're not currently implemented correctly is that I stupidly forgot to sanity check the schema in the issue against the Wiki.

And yep, totally fine splitting the work toward #17  into a separate PR. Will get that done and posted later today.

concert/cloudformation.yml
Borchardt: 😂

Yeah, I meant to add an explanatory comment here. When using `!GetAtt AggregateChoiceDataS3Bucket.Arn`, I couldn't get this policy to properly grant access to the bucket. I'm still not totally sure why that was happening, but it seemed like the workaround was probably to specify the Arn using the `!Sub` pattern used elsewhere above. I thought that that meant I'd probably need to define a bucket name for the S3 resource, and since bucket names are globally unique (right?), I didn't want to add that update to my template and steal our bucket names without getting your take first

sjawhar:
Recapping our conversation. I was making things too complicated. What you have will work great with a few small changes:
* Include colors by setting CHOICE_TYPE to the color value and summing 1
* Change to a stagger window

Borchardt:
NOTE: NOT READY FOR REVIEW IN CURRENT STATE.

concert/cloudformation.choice-aggregator.yml
```diff
@@ -0,0 +1,282 @@
...
+                    CHOICE_TYPE || '_' || CHOICE,
+                    COUNT(CHOICE) OVER (
+                      PARTITION BY CHOICE_TYPE, CHOICE
+                      RANGE INTERVAL '1' MINUTE PRECEDING
+                    ) AS choiceSum,
+                    COUNT(*) OVER W1 AS choiceCount
+                  FROM ${SourceStreamPrefix}_001 AS SOURCE
+                  WHERE CHOICE_TYPE LIKE 'CHOICE_COLOR'
+                  WINDOW W1 AS (
+                    PARTITION BY CHOICE_TIME
```
sjawhar: I think you also need to partition by `CHOICE` for color. Then you can simply choiceSum and choiceCount to just be `COUNT(*) OVER W1`

concert/cloudformation.choice-aggregator.yml
```diff
@@ -0,0 +1,282 @@
...
+                  INSERT INTO ${DestinationStream}
+                  SELECT STREAM CHOICE_TIME,
+                    CHOICE_TYPE || '_' || CHOICE,
+                    COUNT(CHOICE) OVER (
+                      PARTITION BY CHOICE_TYPE, CHOICE
+                      RANGE INTERVAL '1' MINUTE PRECEDING
+                    ) AS choiceSum,
+                    COUNT(*) OVER W1 AS choiceCount
+                  FROM ${SourceStreamPrefix}_001 AS SOURCE
+                  WHERE CHOICE_TYPE LIKE 'CHOICE_COLOR'
```
sjawhar: If you don't have any wildcards, isn't `LIKE 'CHOICE_COLOR'` the same as `= CHOICE_COLOR`?

concert/cloudformation.choice-aggregator.yml
Borchardt: Yep, this is what I ended up doing for the most recent revision after our discussion. At this point, I was still trying to figure out how to set choiceCount to the number of choices for all colors at this timestamp. Out of curiosity, do you know what would have been the right way to make that happen?

concert/cloudformation.choice-aggregator.yml
Borchardt: I'll take your word that it is ;)

Out of curiosity: although Kinesis accepts `CHOICE_TYPE = 'CHOICE_COLOR'`, it seems not to like `CHOICE_TYPE != 'CHOICE_COLOR'`; any idea why that might be? (This is why I've kept `CHOICE_TYPE NOT LIKE 'CHOICE_COLOR'` for the non-color select statement in the most recent revision.)

concert/cloudformation.choice-aggregator.yml
sjawhar: Two different windows, one partitioning by CHOICE and one not

concert/cloudformation.choice-aggregator.yml
sjawhar: Looking at [the docs](https://docs.aws.amazon.com/kinesisanalytics/latest/sqlref/sql-reference-logical-operators.html), it looks like it should be `<>`

concert/cloudformation.choice-aggregator.yml
```diff
@@ -0,0 +1,276 @@
...
+                CREATE OR REPLACE PUMP COUNT_PUMP AS
+                  INSERT INTO ${DestinationStream}
+                  SELECT STREAM CHOICE_TIME,
+                    CHOICE_TYPE || '_' || CHOICE,
+                    COUNT(*) AS choiceSum,
+                    COUNT(*) AS choiceCount
+                  FROM ${SourceStreamPrefix}_001 AS SOURCE
+                  WHERE CHOICE_TYPE = 'CHOICE_COLOR'
+                  WINDOWED BY STAGGER (
+                    PARTITION BY CHOICE_TIME, CHOICE_TYPE, CHOICE
```
sjawhar: CHOICE_TYPE partition here is redundant since it's specified in the WHERE clause.

concert/cloudformation.choice-aggregator.yml
```diff
@@ -0,0 +1,276 @@
...
+                    CHOICE_TYPE,
+                    SUM(CAST(CHOICE AS DOUBLE)) AS choiceSum,
+                    COUNT(*) AS choiceCount
+                  FROM ${SourceStreamPrefix}_001 AS SOURCE
+                  WHERE CHOICE_TYPE NOT LIKE 'CHOICE_COLOR'
+                  WINDOWED BY STAGGER (
+                    PARTITION BY CHOICE_TIME, CHOICE_TYPE
+                    RANGE INTERVAL '1' MINUTE
+                  );
+                CREATE OR REPLACE PUMP COUNT_PUMP AS
```
sjawhar: I think COLOR_PUMP is a clearer name

concert/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
sjawhar: What's this file?

start_kinesis_request.json
```diff
@@ -0,0 +1,13 @@
+{
```
sjawhar: What's this file?

concert/tmp.txt
```diff
@@ -0,0 +1,428 @@
+{
```
sjawhar: What's this file?

concert/app/src/functions/on_choice_made/index.py
```diff
@@ -6,10 +6,10 @@
 Processes websocket API messages from API Gateway
 """
 
-AGGREGATE_CHOICE_DELIVERY_STREAM_NAME = os.environ.get(
+AGGREGATE_CHOICE_DELIVERY_STREAM_NAME = os.environ[
     "AGGREGATE_CHOICE_DELIVERY_STREAM_NAME"
-)
-CHOICE_MADE_SNS_TOPIC_ARN = os.environ.get("CHOICE_MADE_SNS_TOPIC_ARN")
+]
+CHOICE_MADE_SNS_TOPIC_ARN = os.environ["CHOICE_MADE_SNS_TOPIC_ARN"]
```
sjawhar: Why this change? The next one still uses `.get()`

concert/cloudformation.choice-aggregator.yml
```diff
@@ -0,0 +1,276 @@
...
+                  CHOICE_COUNT  int
+                );
+                CREATE OR REPLACE PUMP SUM_PUMP AS
+                  INSERT INTO ${DestinationStream}
+                  SELECT STREAM CHOICE_TIME,
+                    CHOICE_TYPE,
+                    SUM(CAST(CHOICE AS DOUBLE)) AS choiceSum,
+                    COUNT(*) AS choiceCount
+                  FROM ${SourceStreamPrefix}_001 AS SOURCE
+                  WHERE CHOICE_TYPE NOT LIKE 'CHOICE_COLOR'
```
sjawhar: If you wanted to not use LIKE, you could do `<>`

concert/yarn.lock
Borchardt: Committed by mistake. Now removed.

concert/tmp.txt
Borchardt: Committed by mistake. Now removed.

start_kinesis_request.json
Borchardt: This is the CLI input used for starting the Kinesis application via the following command:
`aws kinesisanalyticsv2 start-application --cli-input-json file://start_kinesis_request.json`
If we don't want it, I don't mind deleting. Just thought it might be useful to have.

concert/app/src/functions/on_choice_made/index.py
Borchardt: Yeah, this should still be `.get()`; don't think this was intentional. Fixed. Thanks for catching.

concert/cloudformation.choice-aggregator.yml
Borchardt: IMO, `NOT LIKE` is a little bit more legible. If you're okay with it, will just leave it as is,

start_kinesis_request.json
sjawhar: Ah, that does seem useful! I think you mentioned that the application doesn't start automatically when the cloudformation stack is deployed?

If you wanted to take it to the next level, you could add a [custom lambda-backed resource](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cfn-customresource.html)

concert/cloudformation.choice-aggregator.yml
sjawhar: :+1:

PR #34, opened 2019-07-25
Issue17

### LINKED ISSUE #17: When a user makes a choice, that choice is published to choice-made SNS, Kinesis, and callback SNS

Borchardt:
## Purpose
* Closes #17 
* Split out from #33 

## Approach
* Add Firehose delivery stream and S3 destination
* Lambda handler publishes to choice_made SNS topic, visualization SNS topic, and Firehose delivery stream
* Add tests for `on-choice-made` to assert that both SNS (publish to both choice_made and visualization) and Firehose (put_record) are called upon message receipt

## Testing
* Tests pass
* CloudFormation template deployed successfully
* `on-choice-made` Lambda invoked successfully with test event


concert/cloudformation.yml
```diff
@@ -383,6 +386,65 @@ Resources:
...
+                Action: s3:*
+                Resource: !GetAtt AggregateChoiceDataS3Bucket.Arn
+              - Effect: Allow
+                Action: logs:PutLogEvents
+                Resource: !GetAtt AggregateChoiceDataStreamLogGroup.Arn
+
+  AggregateChoiceDataS3Bucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      BucketName: !Sub ${Environment}-enchanted-brain-choices
```
Borchardt: I'm hoping that my having deployed this template for testing doesn't cause issues with availability of this bucket name. I've deleted my test stack, so I think that we should be good to go, but may need to change this slightly if I'm wrong about that.

PR #44, opened 2019-07-30
Synesthesia visualization

### LINKED ISSUE #24: Aggregated user color and emotion choices are visualized
## Expected Behavior
User choices are going to come streaming in through the websocket connection:
```js
{
  event: 'CHOICE_MADE',
  data: {
    userId: String,
    choiceType: String, // CHOICE_COLOR | CHOICE_EMOTION_HAPPINESS | CHOICE_EMOTION_AGITATION | CHOICE_CHILLS
    choice: String || Number,   // String for color, Number for the rest
    timestamp: String,
  },
}
```

The current idea is a "bees" visualization, where colored dots appear on a 2D plane. The x axis is happy/sad, the y axis is agitated/calm, and the color is... well it's color, obviously.

Not a terrible amount of detail here because we're not quite sure what this should be. So GO WILD!

sjawhar:
## Purpose
Closes #24 

## Approach
* Added simple react app for visualization.
* Uses a ring buffer of configurable size. Three of them actually, one for each dimension. As choices stream in, then are added to the buffer in order.
* Also had to switch authentication source for the API to a querystring parameter. Since the visualization runs in the browser, it can't add custom headers. However, the app still uses the header for authentication (more secure) and just uses a unique but meaningless value for the query string token
* Added a script to simulate audience responses.

## Screenshots
image

### With Backend
* Deploy the backend from this branch
* create `.env.local` under visualization/app with the following line: `REACT_APP_WEBSOCKET_EMITTER_EXPOSE=true`
* `dcdev up` in visualization to launch the visualizer
* Open http://localhost:3000 and log in with a visualizer user
    * username: sami.n.jawhar+test.viz.enchantedbrain@gmail.com
    * password: mv3cygL53yX26SusBasjUNhY5kqvVBCY!
* Open the devtools and make sure the CONNECTED event was received
* Run the following in the browser console
    ```js
    emitter.emit('EVENT_STAGE_CHANGED', {
      stageId: 'STAGE_CHOICE_SYNESTHESIA',
      startTime: new Date().toISOString(),
      choiceTypes: ['CHOICE_COLOR',  'CHOICE_EMOTION_HAPPINESS',  'CHOICE_EMOTION_ENERGY']
    });
    ```
* Grab the queue URL for the created visualization queue from AWS
* `npm install` the packages under visualization/simulator
* Run `node visualization/simulator/index.js -u $QUEUE_URL`
  * Run `--help` to get a full argument list

### Without Backend
* Including the line above, add the following line to `.env.local`: `REACT_APP_WEBSOCKET_API_STUB=true`
* `dcdev up`, http://localhost:3000
* emit the EVENT_STAGE_CHANGED event
* run `emitter.emit('CHOICE_MADE', { choiceType: .... })` as desired

### Performance
- [x] Randomize on push, eliminate as many operations from render() as possible
- [x] Only re-render every 25 choices or 250 ms (whichever comes first)

### Arseny Feedback
- [x] Little brain SVGs instead of circles
- [x] Add emotion axes

### Other
- [x] Add token auth to docs


sjawhar:
I think this one's ready to go, pending feedback from Arseny

visualization/app/.env
```diff
@@ -0,0 +1,10 @@
+REACT_APP_AMPLIFY_INCLUDE_GREETING=false
+REACT_APP_AMPLIFY_REGION=us-east-1
+REACT_APP_AMPLIFY_USER_POOL_ID=us-east-1_nNB31bPdm
+REACT_APP_AMPLIFY_USER_POOL_WEB_CLIENT_ID=4iso70ukf2iloror97ll7b9vq7
+REACT_APP_BUFFER_RING_SIZE=250
+REACT_APP_UPDATE_BATCH_SIZE=25
+REACT_APP_UPDATE_TIMEOUT=250
+REACT_APP_WEBSOCKET_API_STUB=false
+REACT_APP_WEBSOCKET_API_URL=wss://f5t62g9095.execute-api.us-east-1.amazonaws.com/Prod
```
Borchardt: For the region, user_pool and client IDs, and websocket URL, should we (can we?) be pulling these from CFN?

visualization/simulate/index.js
```diff
@@ -0,0 +1,98 @@
+const aws = require('aws-sdk');
```
Borchardt: This is really great for end to end testing: it's already helped me catch at least two bugs. For that purpose, would love to see this simulator support chills as well.

(Non-blocking, obviously. I can work on adding this next week if you want.)

visualization/app/.env
sjawhar: I imagine we could find a way to do that as part of a build process. I'll look into that if we decide to deploy this to S3 instead of just running it locally.

PR #45, opened 2019-08-08
Update event stage record on event stage change

### LINKED ISSUE #11: When the event stage changes, the EVENT_STAGE DB record is updated and a message is published to the callback SNS topic
image

## Expected Behavior
When the EVENT_STAGE_CHANGED route is called, a Lambda function should:
1. Update the EVENT_STAGE record in the database with the new data
2. Publish a message to the callback SNS topic

## Websocket Message Format
```js
{
  event: 'EVENT_STAGE_CHANGED',
  data: {
    stageId: String,
    ... // other attributes
  },
}
```

### DB Record Format
```js
{
  recordId: 'EVENT_STAGE',
  stageId: String,
  .. // other attributes
}
```

### SNS Message Payload
Copy of websocket message.

## Context
We will use this to indicate progress the app through the event automatically, prompting the user for the correct input when necessary. For example, when it's time to prompt the user for color choices, the message might look like:
```js
{
  event: 'EVENT_STAGE_CHANGED',
  data: {
    stageId: 'STAGE_CHOICE_COLORS',
    startTime: String,
    endTime: String,
    frequency: Number,
  },
}
```

## Tasks
- [ ] Tests
- [ ] Code
- [ ] CloudFormation template
- [ ] Docs

Borchardt:
## Purpose
Closes #11 

## Approach
* Updates Dynamo `STAGE_ID` record upon receiving `EVENT_STAGE_CHANGED` event
* As currently implemented, `stageId` is the only fixed event field. Any other fields present in an event should still be processed, but if we have an exhaustive list of possible event fields, we may want to lock updates down to only those fields.

## Testing
- [X] Tests pass
- [X] Functional testing: invoking Lambda with event
```
 {"body":"{\"event\":\"EVENT_STAGE_CHANGED\",\"data\":{\"stageId\":\"STAGE_CHOICE_COLOR_EMOTION\",\"displayName\":\"song title\",\"startTime\":\"2019-05-14T21:20:03.000Z\",\"endTime\":\"2019-05-14T21:21:33.000Z\"}}"}
```
results in the following Dynamo row:
image

- [ ] Do the docs need to be updated? Only if we've finalized `EVENT_STAGE_CHANGED` shapes


concert/app/test/on_event_stage_changed/test_on_event_stage_changed.py
```diff
@@ -0,0 +1,104 @@
...
+                "frequency": otherAttributes["frequency"],
+            },
+        }
+    )
+
+    sns_expected_params = {
+        "TopicArn": CALLBACK_SNS_ARN,
+        "Message": expectedMessage,
+        "MessageStructure": "string",
+        "MessageAttributes": {
```
sjawhar: We actually shouldn't be using MessageAttributes anymore. That was based on using topic filters, which we discovered we could not do because of account limits.

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -19,4 +32,26 @@ def handler(event, context):
         },
     )
 
+    stage_record_update_response = update_stage_record(message)
```
sjawhar: What do you think about putting the DynamoDB update first, so that the SNS message isn't sent until the DB is updated properly? If the DB update fails, it'll keep retrying and message won't be sent until success.

concert/app/src/layer/python/enchanted_brain/attributes.py
```diff
@@ -1,5 +1,6 @@
 ATTR_RECORD_ID = "recordId"
 ATTR_CREATED_AT = "createdAt"
+ATTR_STAGE_ID = "stageId"
```
sjawhar: Please call this `ATTR_EVENT_STAGE_ID` to be consistent with record-type-specific attribute names.

concert/app/test/on_event_stage_changed/test_on_event_stage_changed.py
Borchardt: Removed.

concert/app/src/layer/python/enchanted_brain/attributes.py

concert/app/src/functions/on_event_stage_changed/index.py
Borchardt: Makes a lot of sense. Done.

Issue #38, opened 2019-07-29
Audience chills are aggregated


sjawhar:
## Context
Remember that a user reports chills in an event like the following:
```js
{
  event: 'CHOICE_MADE',
  data: {
    choiceType: 'CHOICE_CHILLS',
    choice: Array(Number),         // e.g. [0.1, 0.3, 0.4, 0.6, 0.2]
    timestamp: String,             // Time of touch start
    interval: Number,              // Time between choices, e.g. 0.2 for 200ms
  },
}
```

## Current Behavior
Since the user can start a choice at any time, our current aggregation method has a few flaws:
* The most obvious is that the method expects choice to be a single string or number, not a set of numbers.
* Aggregating over choiceTime is likely to only capture a single user per aggregation window
* If we try to address the above problem with something like `STEP(CHOICE_TIME BY SECOND)`, we'll end up over-counting user choices since there can be multiple choices each second (e.g. 5 per second if the polling frequency is 200ms)

## Possible Solution
I see a few different ways of handling this:
* OnChoiceMade could flatten out the record into multiple records, each with a single choice value and different timestamp, before sending into the Kinesis firehose stream.
* A [processor lambda on the Kinesis Firehose](https://docs.aws.amazon.com/firehose/latest/dev/data-transformation.html) could handle the above instead.
* Neither of the above would fix the timestamp uniqueness issue. Maybe we could have whichever function handles the record flattening also round each timestamp to the nearest 200ms (or whatever the interval is)?

sjawhar:
No longer needed. The client will send multiple CHOICE_MADE events instead.

PR #46, opened 2019-08-10
Aggregate choices are written to DynamoDb

### LINKED ISSUE #10: Aggregated user choices are written to DynamoDB
image

## Expected Behavior
The output of the Kinesis Data Application (#5) should be written to DynamoDB.

### DB Record Format
```js
{
  recordId: 'AGGREGATE',
  colors: {
    [timestamp]: {
      COLOR_BLUE: Number,
      COLOR_GREEN: Number,
      COLOR_RED: Number,
      ...
    },
    ...
  },
  emotions: {
    [timestamp]: {
      EMOTION_HAPPINESS: Number,
      EMOTION_AGITATION: Number,
    },
    ...
  },
  chills: {
    [timestamp]: Number,
  },
}
```

## Context
Aggregated data will be used to provide a visualization of user choices in the app during the intermission.

## Possible Solution
Imagine a user chooses COLOR_RED at timestamp 1568916000. Imagine also that this is the first choice with that timestamp. Then colors.1568916000 will not yet exist, so you can't do `SET colors.1568916000.COLOR_RED = if_not_exists(colors.1568916000.COLOR_RED, 0) + 1`. I see two possible solutions:

1. A transaction of two updates. The first creates the map colors.1568916000, and the second does the above update.
2. The above update with the condition `attribute_exists(colors.1568916000)`. If the update succeeds, we're done. Otherwise, `SET colors.1568916000 = {}` with the condition `attribute_not_exists(colors.1568916000)` (in case another function created it in the meantime). Regardless of this second condition check, repeat the first update call.

Method 1 always has two updates with one network call. Method 2 either has 1 or 3 updates with as many network calls. If there are 750 respondents making choices every 20 seconds or so, that's almost 40 responses per second and only a couple of them should make 3 updates, most should be only one.

## Tasks
- [ ] Tests
- [ ] Code
- [ ] Update CloudFormation template

Borchardt:
## Purpose
Closes #10 

## Approach
* write logic for chills choices may need to be updated after #38 

## Testing
* invoking `on-choice-made` with test event 
```
{
  "body": "{\"data\":{\"choiceType\":\"CHOICE_EMOTION_ENERGY\",\"choice\":1.5,\"timestamp\":\"2019-05-14T21:20:03.000Z\"}}",
  "requestContext": {
    "authorizer": {
      "principalId": "user_id"
    }
  }
}
``` 
results in DynamoDB record
```
recordId: AGGREGATE
chills: {}
colors: {
    2019-05-14 21:20:03.000: {
        EMOTION_ENERGY: 1.5
    }
}
emotions: {}
```

- [X] Do the docs need to be updated? Yes: 
* new aggregate storage shape (DONE)
*  Dynamo needs empty maps for `colors`, `emotions`, and `chills` for `AGGREGATE` row (#50 created to track this)
- [X] Depends on Issue 5


concert/app/src/functions/aggregate_writer/index.py
```diff
@@ -1,2 +1,106 @@
...
+        "ExpressionAttributeValues": {":choice_average": choice_average},
+        "ReturnValues": "NONE",
+    }
+
+    if choice_type.startswith(CHOICE_COLOR):
+        color = choice_type.split("_")[2]
+        choice_type = CHOICE_COLOR
+        update_args[
+            "UpdateExpression"
+        ] = "ADD #choice_key.#timestamp.#color :choice_average"
```
Borchardt: This isn't actually going to work for updates since averages aren't additive. If we care about this, it seems like the best way to allow for updates might be to change the schema for each choice type to something like:
```
colors: {
    [timestamp]: {
        TOTAL_CHOICES: 5
        COLOR_RED: 3
        COLOR_BLUE: 2
    }
}
```

And then handle the averaging in the vizualizer. Thoughts?

concert/app/src/layer/python/enchanted_brain/attributes.py
```diff
@@ -1,5 +1,6 @@
 ATTR_RECORD_ID = "recordId"
 ATTR_CREATED_AT = "createdAt"
+ATTR_TOTAL_CHOICES = "choices"
```
sjawhar: Please call this `ATTR_AGGREGATE_TOTAL_CHOICES` to be consistent with record-type-specific attributes.

concert/app/src/functions/aggregate_writer/index.py
```diff
@@ -1,2 +1,116 @@
...
+
+    else:
+        update_args["ExpressionAttributeNames"][
+            "#choice_string"
+        ] = ATTR_CHOICE_VALUE_CHILLS
+
+    choice_key = CHOICE_TYPE_KEYS[choice_type]
+    update_args["ExpressionAttributeNames"]["#choice_key"] = choice_key
+
+    ensure_map_available_for_record = create_map_for_record_if_none_exists(
```
sjawhar: For performance, this would work better as the first step in a transaction. Only one network call.
```
SET #choiceKey.#timestep = if_not_exists(#choiceKey.#timestep, :emptyMap)
```

concert/app/src/functions/aggregate_writer/index.py
```diff
@@ -1,2 +1,116 @@
...
+
+    if choice_type.startswith(CHOICE_COLOR):
+        color = choice_type.split("_")[2]
+        choice_type = CHOICE_COLOR
+        update_args["ExpressionAttributeNames"]["#choice_string"] = color
+
+    elif choice_type.startswith("CHOICE_EMOTION"):
+        emotion = choice_type[7:]
+        update_args["ExpressionAttributeNames"]["#choice_string"] = emotion
+        update_args["ExpressionAttributeNames"]["#total_choices"] += "_{}".format(
```
sjawhar: To make sure I understand,  is the intention here to create a record like this?
```js
{
  recordId: 'AGGREGATE',
  colors: {
    '2019-08-14T10:07:00.000Z': {
      choices: 34,
      '#AB0000': 12,
      '#00AB00': 9,
      ...
    },
    '2019-08-14T10:07:20.000Z': {
      choices: 35,
      '#AB0000': 11,
      '#00AB00': 10,
      ...
    },
    ...
  },
  emotions: {
    '2019-08-14T10:11:20.000Z': {
      choices_EMOTION_HAPPINESS: 82,
      EMOTION_HAPPINESS: 47.4,
      choices_EMOTION_ENERGY: 76,
      EMOTION_ENERGY: 21.8,
    },
    '2019-08-14T10:11:40.000Z': {
      choices_EMOTION_HAPPINESS: 87,
      EMOTION_HAPPINESS: 51.0,
      choices_EMOTION_ENERGY: 74,
      EMOTION_ENERGY: 20.2,
    },
  },
  chills: {
    '2019-08-14T10:15:40.000Z': {
      chills: 10.4,
      choices: 41,
    },
  },
}
```

I think words like "sum" and "count" and more explicit than generic "choices" for expressing the meaning of an attribute.

BTW it took me around 10 minutes of reading the code to decipher this. `#choice_string` and `#total_choices` might better be called `#choiceSum` and `#choiceCount` to help your fellow developer understand the intention. And please update the docs.

concert/app/test/aggregate_writer/test_aggregate_writer.py
```diff
@@ -0,0 +1,151 @@
...
+
+    elif choice_type.startswith("CHOICE_EMOTION"):
+        expected_aggregate_update_params["ExpressionAttributeNames"][
+            "#choice_string"
+        ] = emotion_or_color
+        expected_aggregate_update_params["ExpressionAttributeNames"][
+            "#total_choices"
+        ] = "choices_{}".format(emotion_or_color)
+
+    elif choice_type == "CHOICE_CHILLS":
```
sjawhar: All we've done here is copy-pasted the code from src to test. That doesn't really test anything. :laughing: 

We should be using static expectation expressions.

concert/app/test/aggregate_writer/test_aggregate_writer.py
```diff
@@ -0,0 +1,151 @@
...
+        ] = "chills"
+
+    resp = None
+    with Stubber(dynamodb.meta.client) as stub:
+        stub.add_response(
+            "update_item",
+            dynamo_update_item_success_response,
+            expected_map_creation_params,
+        )
+        stub.add_response(
```
sjawhar: You shouldn't need to test every service call is each test. Each test should test a single thing.

concert/app/src/layer/python/enchanted_brain/attributes.py

concert/app/src/functions/aggregate_writer/index.py
Borchardt: As I understand it, transactions won't work here, as no item can be targeted by more than one action within a single transaction. Am I mistaken there? Or maybe I'm misunderstanding what you mean?

sjawhar: Nope, my understanding of transactions was faulty. You are correct.

concert/app/src/functions/aggregate_writer/index.py
Borchardt: Sorry for the confusion here; I should've included a screenshot of the output.

I've made the changes you mentioned above. Output is pictured below.

image

concert/app/test/aggregate_writer/test_aggregate_writer.py
Borchardt: Fair ask. I've tried to do a better job of this in the most recent revision.

concert/app/test/aggregate_writer/test_aggregate_writer.py
Borchardt: Fair ask. I've tried to do a better job of this in the most recent revision.

PR #48, opened 2019-08-14
update song list on event stage change

### LINKED ISSUE #39: When the event stage changes, the SONG_LIST record is updated
## Expected Behavior
This builds off of the behavior of #11. Remember that, at the [end of the app-guided part of the event](https://github.com/sjawhar/enchanted-brain/wiki/Concert#5---intermission-and-end), the user will be sent the following event:
```js
{
  stageId: 'STAGE_END',
  songs: [
    {
      displayName: String,
      startTime: String,
      endTime: String,
    },
    ...
  ],
  colors: [...],
  chills: [...],
}
```

Remember also that the song list is stored in a record as follows
```js
{
  recordId: 'SONG_LIST'
  songs: [
    {
      displayName: String,  // Helpful for in-app displays
      startTime: String,
      endTime: String,
    },
    ...
  ],
}
```

Finally, remember from #11 that event stage changes are signaled with `EVENT_STAGE_CHANGED` event, which include such information as `startTime` and `endTime`

It therefore seems logical to build the SONG_LIST record automatically as the event goes along by recording the startTime and endTime of the passing events.

## Context
The song list will be used at the end of the app-guided part of the event to show the user aggregated audience choices for each song.

## Possible Solution
Maybe we need to add `displayName` to the `EVENT_STAGE_CHANGED` events, for this functionality to make use of. If so, just update the docs accordingly.

Borchardt:
## Purpose
Closes #39 

## Testing
- [X] Tests pass
- [X] Functional testing completed: invoking Lambda with test event 
```
 {"body":"{\"event\":\"EVENT_STAGE_CHANGED\",\"data\":{\"stageId\":\"STAGE_CHOICE_COLOR_EMOTION\",\"displayName\":\"song title\",\"startTime\":\"2019-05-14T21:20:03.000Z\",\"endTime\":\"2019-05-14T21:21:33.000Z\"}}"}
``` 
results in the following Dynamo record:
image

- [X] Do the docs need to be updated?


concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -19,4 +35,36 @@ def handler(event, context):
...
+
+
+def update_song_list(message):
+    response = None
+
+    data = message["data"]
+    display_name = data.get(ATTR_DISPLAY_NAME)
+    start_time = data.get(ATTR_START_TIME)
+    end_time = data.get(ATTR_END_TIME)
+    if display_name and start_time and end_time:
```
sjawhar: What about if not?

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -19,4 +35,36 @@ def handler(event, context):
         },
     )
 
+    song_list_update_response = update_song_list(message)
```
sjawhar: Neither this variable nor `sns_response` are being used.

concert/app/src/layer/python/enchanted_brain/attributes.py
```diff
@@ -1,5 +1,10 @@
 ATTR_RECORD_ID = "recordId"
 ATTR_CREATED_AT = "createdAt"
+ATTR_STAGE_ID = "stageId"
+ATTR_DISPLAY_NAME = "displayName"
+ATTR_END_TIME = "endTime"
+ATTR_SONGS = "songs"
```
sjawhar: Please use prefixes (e.g. `ATTR_SONG_LIST_*`) for attributes that are only used for one record type.

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -19,4 +35,36 @@ def handler(event, context):
...
+
+        update_args = {
+            "Key": {ATTR_RECORD_ID: RECORD_ID_SONG_LIST},
+            "UpdateExpression": "SET #songs = list_append(#songs, :song)",
+            "ExpressionAttributeNames": {"#songs": ATTR_SONGS},
+            "ExpressionAttributeValues": {":song": song},
+            "ReturnValues": "NONE",
+        }
+
+        response = table.update_item(**update_args)
```
sjawhar: Since this issue builds off of #11, it makes sense to me to do all these updates in a single transaction. One network call.

docs/Usage/Concert/Concert.md
```diff
@@ -71,6 +71,7 @@ The user should be displayed the survey form found at `formUrl`, where they will
 ```js
 {
   stageId: 'STAGE_CHOICE_SYNESTHESIA',
+  displayName: String,
```
sjawhar: Thank you for updating the docs.

concert/app/src/functions/on_event_stage_changed/index.py
Borchardt: Intent here is to do nothing if not. My understanding is that some `EVENT_STAGE_CHANGED` events -- e.g. mental imagery, intermission -- are not expected to contain any song list information, so for such events, just skip the update.

sjawhar: Wouldn't this be a more explicit way of expressing that idea?
```suggestion
    if not (display_name and not start_time and end_time):
        return
    # Continue with function login
```

Early returns are your friend.

concert/app/src/functions/on_event_stage_changed/index.py
Borchardt: Done. Will have revision posted once I've finished updating tests.

I'm wondering, though: would it make more sense for the stageId write to be a `PutItem` rather than an `UpdateItem` request? Do we want attributes from old stages to linger?

concert/app/src/layer/python/enchanted_brain/attributes.py

concert/app/src/functions/on_event_stage_changed/index.py
sjawhar: Yes, I think we would want a PutItem instead. Also, my suggestion about transactions might be a poor one considering my misunderstanding about updating the same record multiple times, as you pointed out.

Borchardt: Since `SONG_LIST` and `EVENT_STAGE` are different records, and since each is only updated once per invocation of this handler, I think transactions should be fine here.

(Their syntax is really ugly, though, so I'd be totally fine not using them :wink: )

concert/app/src/functions/on_event_stage_changed/index.py
Borchardt: Done.

concert/app/src/functions/on_event_stage_changed/index.py
Borchardt: Yep, just seemed like we were following a pattern within the handlers of `service_response = serviceCall()`. Removed.

Borchardt:
In response to your question above:

This update will fail if the `SONG_LIST` record doesn't exist or if it doesn't contain the `songs` list attribute. I've added creation of this record and that attribute to #50.

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -35,21 +34,65 @@ def handler(event, context):
...
+        },
+        "TableName": DYNAMODB_TABLE_NAME,
     }
 
-    for key, value in data.items():
-        update_args["UpdateExpression"] += ", #{} = :{}".format(key, key)
-        update_args["ExpressionAttributeNames"]["#{}".format(key)] = key
-        update_args["ExpressionAttributeValues"][":{}".format(key)] = value
+    for key, value in message_data.items():
+        if isinstance(value, int):
```
Borchardt: I don't believe that we have any non-int or string values planned for the event stage record, but if we do, we may need to update this type check.

sjawhar:
> In response to your question above:
> 
> This update will fail if the `SONG_LIST` record doesn't exist or if it doesn't contain the `songs` list attribute. I've added creation of this record and that attribute to #50.

Not if you do this:
```py
    song_list_update_args = {
        "Key": {ATTR_RECORD_ID: {"S": RECORD_ID_SONG_LIST}},
        "UpdateExpression": "SET #songs = list_append(if_not_exists(#songs, :empty_list), :song)",
        "ExpressionAttributeNames": {"#songs": ATTR_SONG_LIST_SONGS},
        "ExpressionAttributeValues": {
            ":song": {"L": song},
            ":empty_list": {"L": []}
        },
        "TableName": DYNAMODB_TABLE_NAME,
    }
```

Borchardt:
Ah, clever. Done, and tested successfully.

concert/app/src/functions/on_event_stage_changed/index.py
sjawhar: You can do this using boto3:
```py
serializer = boto3.dynamodb.types.TypeSerializer()
event_stage_update_args["Item"] = {k: serializer.serialize(v) for k,v in message_data.items()}
```

concert/app/test/on_event_stage_changed/test_on_event_stage_changed.py
```diff
@@ -11,83 +11,232 @@
...
+        data["displayName"] = display_name
+    if start_time:
+        data["startTime"] = start_time
+    if end_time:
+        data["endTime"] = end_time
+    if frequency:
+        data["frequency"] = frequency
+    if form_url:
+        data["formUrl"] = form_url
+    return {"body": json.dumps({"event": "EVENT_STAGE_CHANGED", "data": data})}
```
sjawhar: From what I can tell, `event["data"]` isn't being used. Is that right?

Borchardt: Not sure I understand what you're asking, so going to answer both ways I could see the question being intended.

If asking about whether the shape here matches the shape of the event that the Lambda will receive as input:
Yes, I think so. The websocket message gets passed in the `"body"` field of the event sent to the Lambda by API Gateway, so the fields we care about are at `json.loads(event["body"])["data"]` rather than `event["data"]`.

If asking about whether these fields are used for the tests:
The `test_event_stage_and_song_list_updated_on_event_stage_change` test case tests that the values in the `data` dict created here are properly written to Dynamo. (The `test_callback_message_published_on_event_stage_change` doesn't do any assertions on static content, but I can go ahead and update it to do so.)

If asking about some other thing:
Sorry for the misunderstanding, and please let me know!

concert/app/src/functions/on_event_stage_changed/index.py
Borchardt: Good suggestion. I feel much better about that. Done.

concert/app/test/on_event_stage_changed/test_on_event_stage_changed.py
sjawhar: I was misreading the code. I thought the `event` dict had both a `body` and `data` key, rather than `data` being nested under `body`. My bad.

PR #49, opened 2019-08-15
Chills selector

### LINKED ISSUE #37: A user reports feeling chills
## Expected Behavior
```js
{
  stageId: 'STAGE_CHOICE_CHILLS',
  startTime: String,
  endTime: String,
  interval: Number,
}
```
* When the above stage is reached, the user should be taken to the chills selector
* Centered in the screen should be a touchable box. This box should take up most of the height of the screen, but not the entire width. The point is that it should be clear to the user that this is an interactable input.
* Touching anywhere within this box should start recording chills. The further up the box the user touches, the higher the value of reported chills. The top is 1, bottom is 0. The value should be recorded every `interval` seconds.
* Not touching = stop recording and send the data to the server

For each recorded value, the client should send an event as follows:
```js
{
  event: 'CHOICE_MADE',
  data: {
    choiceType: 'CHOICE_CHILLS',
    choice: Number,      // between 0 and 1
    timestamp: String,   // Time of recorded touch
  },
}
```

## Context
If it's too hard to do a box instead of the entire screen, I guess the entire screen would work. I just think it would be less clear, and we'd also have some noise around the top and bottom based on different phones.

## Possible Solution
I think something like this should work:
* On touch, `setInterval()` with some polling frequency (e.g. every 200 ms). Record the start time. On each poll, push the current value to an array.
* On touch end, `cancelInterval()` and send the array of touch points.

We could also then use this array to visualize the user choice as it's being made as a kind of waveform scrolling to the left of the input box:

### Tasks
- [x] Touch input box
- [x] Waveform
- [x] Waveform disappear on release
- [x] Send data

sjawhar:
## Purpose
Closes #37 

## Approach
Pretty much what is described in the issue. Also embedded some instructions in the touch input box that don't interfere with touching. Some notes:
* Choice value is rounded to 2 decimal places (e.g. 0.12, 0.75) before sending
* We can control the number of polled values shown on the screen by changing `INPUT_BUFFER` (hard-coded value)
* We can control the polling frequency through `interval` (value sent by API)
* The waveform fades away after letting off

## Testing
```bash
cat << EOF | git apply
--- a/client/app/src/api/concertApi.js
+++ b/client/app/src/api/concertApi.js
@@ -6,6 +6,15 @@ let isConnect = false;
 const events = new EventEmitter();
 
 const connect = idToken => {
+  const now = new Date();
+  events.emit('EVENT_STAGE_CHANGED', {
+    stageId: 'STAGE_CHOICE_CHILLS',
+    startTime: now.toISOString(),
+    endTime: new Date(now.setMinutes(now.getMinutes() + 2)).toISOString(),
+    interval: 0.2,
+  });
+  return;
+
   if (ws) {
     return;
   }
@@ -53,6 +62,7 @@ const disconnect = () => {
 };
 
 const send = message => {
+  console.log('SEND', message);
   if (!ws) {
     return false;
   }
EOF
```

- [x] Confirm with @Borchardt that aggregation will work if timestamps are rounded to 200ms
- [x] Get feedback from Arseny


Borchardt:
Re: rounding to 200ms and aggregation:

We're not using Kinesis's floor function anymore, so I think that the timestamps are effectively just being handled as strings. 200ms intervals should be fine. I've gone ahead and tested with a few manual Lambda invocations which produced the output below:

image

sjawhar:
> Looks good! It's a little bit laggy on my phone, but imagine that that might have something to do with building and running through Expo. Will reserve judgment there until I've seen the production build.

Not just your phone. The screen only updates every 200ms. I'm not a fan either, and I've had an idea of how to improve. Will push an update soon.

sjawhar:
@Borchardt can you take another look. I made it feel less laggy.

Borchardt:
Definitely more smooth. Good work!

One odd thing I notice with the new version is that when I stop selecting, all previous selections are cleared from the screen. Is that intentional?

Being nitpicky, the waveform starts to look more like a scatter plot if I move my finger a little bit quickly. Do you think that that's something we care about? If so, it seemed like the rectangular markers used in the first version helped a little bit both with this and with smoothing the peaks of the waveform. Using connecting lines between marker points might be another solution.

I see all of these as "nice to haves" rather than as blockers, so approving.

sjawhar:
> One odd thing I notice with the new version is that when I stop selecting, all previous selections are cleared from the screen. Is that intentional?

It was always supposed to be doing that, so if it wasn't before then it's working better now. Yes, the intention is for them to fade away.

> Being nitpicky, the waveform starts to look more like a scatter plot if I move my finger a little bit quickly. Do you think that that's something we care about? If so, it seemed like the rectangular markers used in the first version helped a little bit both with this and with smoothing the peaks of the waveform. Using connecting lines between marker points might be another solution.

I thought the rectangular markers looked like I'm playing pong or breakout :laughing: Still waiting for feedback from Arseny. I might look into using a [line chart](https://github.com/JesperLekland/react-native-svg-charts) instead, but right now my focus is on the results screen.

PR #52, opened 2019-08-22
Results Screen

### LINKED ISSUE #43: On STAGE_END, the websocket connection is closed and aggregated audience choices are visualized
## Expected Behavior
```
{
  stageId: 'STAGE_END',
  songs: [
    {
      displayName: String,
      startTime: String,
      endTime: String,
    },
    ...
  ],
  colors: [
    {
      timestamp: String,
      choices: {
        COLOR_BLUE: Number,
        COLOR_GREEN: Number,
        COLOR_RED: Number,
        ...
      },
    },
    ...
  ],
  chills: [
   {
     timestamp: String,
     chills: Number,
   },
   ...
  ],
}
```
This indicates the end of the part of the concert that uses the app. At this point, the aggregated responses/visualizations should be made available in the app and the websocket connection can be closed.

`songs` contains the names, start times, and end times of each of the performed pieces. By storing a user's responses locally, the app can assign them to the appropriate song and display them back to the user. In addition, aggregated choices are available under `colors` and `chills` and can be visualized in the app.

Make sure to check on the status of #40 in case the STAGE_END record structure needs to change.

## Tasks
- [ ] Ensure user choices are being saved locally throughout the event for display in this section
- [ ] Visualize colors
- [ ] Visualize chills

sjawhar:
## Purpose
Closes #43

## Approach
* Stacked area chart for colors, with bar charts below for user choices
* Area chart for chills, with overlayed line chart for user choices
* All displayed in a carousel
* Moved concertApi interaction to store, and store now saves user choices
* Added a stub file to help with testing
* Deleted unused files

## Screenshots
Screenshot_20190822-131107
Screenshot_20190822-131115

## Testing
Set `WEBSOCKET_API_STUB=STAGE_END`  to .env and launch. You might need to close/reopen the app or delete/recreate the docker container to get rid of any cached data.

- [ ] Get feedback from Arseny


Borchardt:
This looks beautiful! One small thing: on some of my color results, the top of the color chart is clipped. See below screenshot for an example of what I mean.

image

sjawhar:
Thanks for catching that. I've pushed a fix. Let me know if you still see it

Borchardt:
Less clipping and less often, but am still seeing it. Two examples:
image
image

sjawhar:
I've handled it better

client/app/src/api/stub.js
```diff
@@ -0,0 +1,151 @@
...
+const colors = swatchColors.flat();
+const now = new Date();
+
+const generateStageEnd = () => {
+  const eventData = {
+    stageId: 'STAGE_END',
+    songs: [
+      {
+        displayName: 'Color Song 1',
+        startTime: new Date(new Date().setMinutes(now.getMinutes() - 34)).toISOString(),
```
Borchardt: JS ignorance question: what's the reason for the wrapped `Date` objects for all of these?

sjawhar: tl;dr: non-fluid JS syntax

The end result I want is an ISO string, which you get by calling `.toISOString()` on a Date object. So you should exect the final form to look something like:
```js
date.toISOString()
```

So, how to get the necessary `date` object?  Well, I want to add and remove some time from a fixed time point (`now`). Can't just do `now + 180`, that returns a meaningless string. I can use `.setMinutes()`, but that modifies `now`. I don't want to modify `now`, because I'm reusing that elsewhere. So I need a new object. That gives us something like
```js
new Date().setMinutes(now.getMinutes() - 3).toISOString()
```

_Except_ `.setMinutes()` returns a number (the unix epoch), not a date object. So I can't chain a call to `.toISOString()` without first wrapping the result in another date object. That gives us

```js
new Date(new Date().setMInutes(now.getMintues() - 3)).toISOString()
```

An alternative would be
```js
const before = new Date();
before.setMinutes(now.getMinutes() - 3);
return before.toISOString();
```

But that's multiple lines. I could do
```js
new Date(now.valueOf() - 180000).toISOString()
```

But I think that's less semantic. In short, I haven't found a better way of doing it using the syntax available in vanilla JS. It's something that's bothered me for a while.

client/app/src/screens/Results.js
```diff
@@ -1,12 +1,244 @@
...
+import Constants from 'expo-constants';
+
+import { store } from '../state';
+import Layout from '../constants/Layout';
+import COLORS, { swatchColorInfo } from '../constants/Colors';
+import { CHOICE_COLOR, CHOICE_CHILLS } from '../constants/Choices';
+
+const COLOR_KEYS = Object.keys(swatchColorInfo);
+const COLOR_EMPTY = Object.fromEntries(COLOR_KEYS.map(color => [color, 0]));
+const CARD_MARGIN = 0.05 * Layout.window.width;
```
Borchardt: Another front end ignorance question: how come none of the widths/heights here take up the full 100%?

sjawhar: Aesthetic choice. I want them to look like cards. If you look at the screenshots and then imagine (or just try it) the margin being gone, I think it would be less intuitive that these are swipeable elements.

client/app/src/screens/Results.js
```diff
@@ -1,12 +1,244 @@
...
+          choiceType,
+          chartProps: {
+            xAccessor: ({ item }) => item.timestamp,
+            xMax: new Date(endTime),
+            xMin: new Date(startTime),
+            xScale: scaleTime,
+            contentInset: {
+              top: Platform.select({
+                ios: 0,
+                android: 0.1 * CHART_COLOR_AGGREGATE_HEIGHT,
```
Borchardt: One last ignorance question for this review: why the different handling here?

sjawhar: > An object that specifies how much fake "margin" to use inside of the SVG canvas. This is particularly helpful on Android where overflow: "visible" isn't supported and might cause clipping.

https://github.com/JesperLekland/react-native-svg-charts/blob/dev/README.md#common-props

PR #54, opened 2019-08-23
Feature/40 stage end includes data

### LINKED ISSUE #40: On STAGE_END, OnEventStageChangedFunction should include song list and aggregated choice data
## Expected Behavior
In the STAGE_END stage data, we need to include aggregated choices and song list for the user's app to visualize. It makes sense to pre-process the data so that each song also includes the aggregated choices that belong to it. In other words, the event should look something like this:

```js
{
  stageId: 'STAGE_END',
  songs: [
    {
      displayName: String,
      startTime: String,
      endTime: String,
      choiceType: String,  // CHOICE_COLOR || CHOICE_CHILLS
      choices: [
        // CHOICE_COLOR
        {
          timestamp: String,
          '#AB0000': Number,
          '#AB8000': Number,
          ...
        },
        // CHOICE_CHILLS
        {
          timestamp: String,
          sum: Number,
          count: Number,
        },
        ...
      ]
    }
    ...
  ]
}
```

`choices` should include elements of _either_ the CHOICE_COLOR or CHOICE_CHILLS format, but not both.

## Possible Solution
It seems to me that a natural place to put this is the OnEventStageChangedFunction. If the stageId is STAGE_END, grab the [AGGREGATE](https://github.com/sjawhar/enchanted-brain/wiki/Concert#aggregated-choices) and [SONG_LIST](https://github.com/sjawhar/enchanted-brain/wiki/Concert#song-information) records and include the necessary data in the new EVENT_STAGE record. That way, the data will be included both in the published message (#11) and if a user disconnects and reconnects during intermission (#13).

## Tasks
- [ ] Update docs
- [ ] Tests
- [ ] Code
- [ ] Check CloudFormation resources and permissions

Borchardt:
## Purpose
Closes #40 

## Screenshots
Sending `STAGE_END` with aggregate choices data as pictured in first screenshot below results in `EVENT_STAGE` record pictured in second screenshot below.
image
image

## Testing
Unit tests pass
Functional testing as described above

- [X] Do the docs need to be updated?


concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -88,3 +104,107 @@ def get_song_list_update_transaction_item(message_data):
...
+    songs_with_aggregate_choices = []
+    for song_metadata in song_list[ATTR_SONG_LIST_SONGS]:
+        display_name = song_metadata[ATTR_SONG_LIST_DISPLAY_NAME]
+        start_time = song_metadata[ATTR_SONG_LIST_START_TIME]
+        end_time = song_metadata[ATTR_SONG_LIST_END_TIME]
+
+        choice_type = None
+
+        for timestamp in aggregate_data[ATTR_CHOICE_VALUE_COLOR].keys():
+            if start_time <= timestamp < end_time:
```
Borchardt: For this data comparison and the others below, we probably want to parse to `datetime` rather than doing string comparisons. Agreed?

sjawhar: If they're ISO format, it doesn't matter.

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -88,3 +104,107 @@ def get_song_list_update_transaction_item(message_data):
     }
 
     return {"Update": song_list_update_args}
+
+
+def get_song_list_get_transaction_item():
+    serializer = boto3.dynamodb.types.TypeSerializer()
```
sjawhar: This could probably be a global object, to reduce memory usage and garbage collection.

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -88,3 +104,107 @@ def get_song_list_update_transaction_item(message_data):
...
+        "TableName": DYNAMODB_TABLE_NAME,
+    }
+
+    return {"Get": song_list_get_args}
+
+
+def get_aggregate_data_get_transaction_item():
+    serializer = boto3.dynamodb.types.TypeSerializer()
+    aggregate_data_get_args = {
+        "Key": {ATTR_RECORD_ID: serializer.serialize(RECORD_ID_AGGREGATE)},
```
sjawhar: These two functions are identical except for a single argument. DRY opportunity.

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -88,3 +104,107 @@ def get_song_list_update_transaction_item(message_data):
...
+    }
+
+    return {"Get": aggregate_data_get_args}
+
+
+def get_song_list_and_aggregate_data():
+    song_list_get_transaction_item = get_song_list_get_transaction_item()
+    aggregate_data_get_transaction_item = get_aggregate_data_get_transaction_item()
+
+    transaction_items = [
```
sjawhar: Style: `song_list_get_transaction_item`, `aggregate_data_get_transaction_item`, and `transaction_items` are each only used once. You could make this easier to read with
```py
responses = dynamodb.transact_get_items(TransactItems=[
    get_transaction_item(RECORD_ID_SONG_LIST),
    get_transaction_item(RECORD_ID_AGGREGATE),
])["Responses"]
```

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -88,3 +104,107 @@ def get_song_list_update_transaction_item(message_data):
...
+        song_list_get_transaction_item,
+        aggregate_data_get_transaction_item,
+    ]
+
+    responses = dynamodb.transact_get_items(TransactItems=transaction_items)[
+        "Responses"
+    ]
+
+    deserializer = boto3.dynamodb.types.TypeDeserializer()
```
sjawhar: Can make this a global object too

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -88,3 +104,107 @@ def get_song_list_update_transaction_item(message_data):
...
+    songs = {k: deserializer.deserialize(v) for k, v in responses[0]["Item"].items()}
+    aggregate_data = {
+        k: deserializer.deserialize(v) for k, v in responses[1]["Item"].items()
```
sjawhar: DRY: Could make this a `deserialize_db_item()` helper function

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -22,11 +33,16 @@
 def handler(event, context):
     message = json.loads(event["body"])
 
+    if message["data"][ATTR_EVENT_STAGE_ID] == EVENT_STAGE_END:
+        song_list, aggregate_data = get_song_list_and_aggregate_data()
+        songs_with_choices = get_songs_with_aggregate_choices(song_list, aggregate_data)
```
sjawhar:
```suggestion
        message["data"][ATTR_SONG_LIST_SONGS] = get_songs_with_aggregate_choices(song_list, aggregate_data)
```

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -88,3 +104,107 @@ def get_song_list_update_transaction_item(message_data):
...
+            choice_type = ATTR_CHOICE_VALUE_CHILLS
+
+        choices = []
+        for timestamp, aggregate_choice_data in aggregate_data[choice_type].items():
+            if start_time <= timestamp < end_time:
+                choices_at_timestamp = {ATTR_CHOICE_TIMESTAMP: timestamp}
+                if choice_type == ATTR_CHOICE_VALUE_COLOR:
+                    choices_at_timestamp.update(
+                        {
+                            k[4:]: v
```
sjawhar: If the aggregator didn't save with the sum_ prefix, lines 170-190 could be replaced with:
```py
choices.append({ ATTR_CHOICE_TIMESTAMP: timestamp, **aggregate_choice_data })
```

concert/app/src/functions/on_event_stage_changed/index.py
Borchardt: Done.

concert/app/src/functions/on_event_stage_changed/index.py
Borchardt: Done.

concert/app/src/functions/on_event_stage_changed/index.py
Borchardt: Done.

concert/app/src/functions/on_event_stage_changed/index.py
Borchardt: Done.

concert/app/src/functions/on_event_stage_changed/index.py
Borchardt: Done.

concert/app/src/functions/on_event_stage_changed/index.py
Borchardt: This isn't quite true, as we'd still need to remove the `count` attribute from the color rows (either here or in the aggregate writer). IMO, that update would introduce as much messiness as we'd clean up here, so I'd prefer to leave this as is. (Though I'm not hugely opposed if you feel strongly.)

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -75,16 +94,104 @@ def get_song_list_update_transaction_item(message_data):
...
+        )
+
+    return songs_with_aggregate_choices
+
+
+def deserialize_db_item(item):
+    return {k: dynamodb_deserializer.deserialize(v) for k, v in item["Item"].items()}
+
+
+def decimal_default(obj):
```
sjawhar: Why reimplement this here instead of just importing `enchanted_brain.parser`?

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -1,32 +1,54 @@
 import boto3
 import json
 import os
+import decimal
```
sjawhar: This isn't being used now

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -1,32 +1,54 @@
...
+from boto3.dynamodb import types
```
sjawhar: The de/serializers are being accessed using `boto3.dynamodb.types`, so this import isn't being used at the moment.

concert/app/src/functions/on_event_stage_changed/index.py
```diff
@@ -1,32 +1,54 @@
...
+        message["data"][ATTR_SONG_LIST_SONGS] = get_songs_with_aggregate_choices(
+            song_list, aggregate_data
+        )
+
     update_event_stage_and_song_list(message)
 
     sns.publish(
         TopicArn=CALLBACK_GLOBAL_SNS_TOPIC_ARN,
-        Message=json.dumps(message),
+        Message=json.dumps(message, default=dynamodb_encoder.default),
```
sjawhar: In OnConnectFunction, it's `json.dumps(message, cls=DynamoDbEncoder)`. Do they both work?

PR #68, opened 2019-08-31
Signup revisions

### LINKED ISSUE #65: Sign Up Revisions
## Expected Behavior
- [x] Make agreement to scientific data use optional, record on user
- [x] Add country of residence, remove name
- [x] Move email and password to separate page

sjawhar:
## Purpose
Closes #65

## Approach
* Broke the signup form into four separate components: Terms, Demographics, User, and UserExistsModal.
* Separated Demographics and User fields into separate steps in the form
* Removed name and added country of residence to demographics
* Accepting scientific use of data is now optional and recorded on the user record

## Screenshots
Pretty much looks identical

## Pre-Merge TODOs
- [ ] Switch to phone number and SMS signup instead of email


users/cloudformation.yml
```diff
@@ -14,21 +14,63 @@ Parameters:
     Type: String
 
 
+Mappings:
+
+  SsmParameters:
```
Borchardt: Minor: sms (I'm fine leaving this as is)

sjawhar: An SSM Parameter is a parameter in the Simple Systems Management parameter store

Borchardt: I stand corrected

PR #72, opened 2019-09-06
custom lambda resource for resource initialization

### LINKED ISSUE #50: Ensure necessary Dynamo items and attributes created on CFN stack deploy
TODO: Compile list of known requirements:
* empty maps for `colors`, `emotions`, and `chills` for `AGGREGATE` row
* `SONG_LIST` record with `songs` list

## Expected Behavior
Records and attributes necessary for writer Lambdas to function successfully created (if don't exist) on CFN deploy

## Current Behavior
These records and attributes have been added manually

## Possible Solution
TODO

## Tasks
TODO

Borchardt:
## Purpose
Closes #50 

## Approach
Add a custom Lambda resource which is triggered on stack changes. This Lambda does the following:
1: Create `colors`, `emotions`, and `chills` maps in the `AGGREGATE` record if they don't exist
2: Start the Kinesis Data Analytics application

## Screenshots
Dynamo after deploying from scratch:

image

Kinesis Data Analytics after deploying from scratch:

image

## Testing
Deployed successfully (see output screenshots above)


concert/app/src/functions/prepare_stack/index.py
```diff
@@ -0,0 +1,95 @@
...
+kinesis_analytics = boto3.client("kinesisanalyticsv2")
+
+
+def handler(event, context):
+    status = "FAILED"
+    try:
+        prepare_aggregate_choices_record()
+        start_kinesis_analytics_application()
+        status = "SUCCESS"
+    finally:
```
sjawhar: It looks like we're swallowing error messages here. Won't the try/finally stop the function from erroring out (which is good), meaning that the error's message won't actually get logged to CloudWatch (which is bad)?

concert/app/test/prepare_stack/test_prepare_stack.py
```diff
@@ -0,0 +1,135 @@
...
+        kinesis_analytics_stub.add_response(
+            "start_application", success_response, expected_params
+        )
+        resp = handler(event, context)
+        kinesis_analytics_stub.assert_no_pending_responses()
+
+    assert resp == {"statusCode": 204}
+
+
+@mock.patch("botocore.vendored.requests.put", return_value=success_response)
```
sjawhar: Love it!

concert/cloudformation.yml
```diff
@@ -361,6 +361,40 @@ Resources:
         FunctionArn: !Ref OnEventStageChangedFunction.Alias
         RouteKey: !FindInMap [RouteKeys, EventStageChanged, Key]
 
+  OnStackChange:
```
sjawhar: Can we move this below PrepareStackFunction, since it references it?

concert/app/src/functions/prepare_stack/index.py
Borchardt: If I'm understanding [Python's try/finally behavior](https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions) correctly, the error will still be raised after completion of the `finally` block, so it should still get logged to CloudWatch

concert/app/src/functions/prepare_stack/index.py
```diff
@@ -0,0 +1,95 @@
...
+        ExpressionAttributeNames={
+            "#chills_map": ATTR_CHOICE_VALUE_CHILLS,
+            "#colors_map": ATTR_CHOICE_VALUE_COLOR,
+            "#emotions_map": ATTR_CHOICE_VALUE_EMOTION,
+        },
+        ExpressionAttributeValues={":empty_map": {}},
+    )
+
+
+def update_aggregate_record(**update_args):
```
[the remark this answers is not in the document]
Borchardt: Fair point. Fixed.

concert/cloudformation.yml
Borchardt: Yep. Done.

concert/app/src/functions/prepare_stack/index.py
sjawhar: Ah, I was wrong about how the try/finally works. But that means this function can error out instead of reporting failure. Is that what we want?

Borchardt: IMO it's fine. As far as I can tell, CloudFormation doesn't know or care how the Lambda terminates or what it returns. Termination status is reported instead through the S3 upload that happens in `put_response` within the `finally` block.

(From [AWS's documentation on custom resources](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html): "Custom resources must send responses to a pre-signed Amazon S3 URL. If they can't send responses to Amazon S3, AWS CloudFormation won't receive a response and the stack operation fails.")

sjawhar: But that's exactly my point. Instead of a timely and meaningful failure reason reported to the S3 url, cloudformation will just timeout waiting for a response that never comes.

sjawhar: Something as simple as this would handle it:
```suggestion
    except Exception as e:
        print("ERROR", e.message)
    put_response(event, context, status)
```

sjawhar: Once again, I was overthinking things. The finally block takes care of reporting back to CloudFormation, so erroring out after that is fine.

concert/cloudformation.yml
```diff
@@ -371,6 +371,39 @@ Resources:
...
+              Resource: !Sub
+                - arn:aws:kinesisanalytics:${AWS::Region}:${AWS::AccountId}:application/${KinesisAnalyticsApplicationName}
+                - { KinesisAnalyticsApplicationName: !GetAtt ChoiceAggregator.Outputs.KinesisAnalyticsApplicationName }
+
+  OnStackChange:
+    Type: Custom::InvokeLambda
+    Condition: IsProduction
+    Properties:
+      ServiceToken: !Ref PrepareStackFunction.Alias
+      StackName: !Sub ${Environment}-enchanted-brain
```
sjawhar: As far as I can tell, this parameter isn't being used

concert/cloudformation.yml
```diff
@@ -371,6 +371,39 @@ Resources:
...
+      DependsOn:
```
sjawhar: This should be a top-level attribute, not nested under `Properties`

Borchardt: Done

concert/cloudformation.yml
Borchardt: Yep, this is an artifact of the tutorial I was using for reference. Good catch. Removed.

PR #73, opened 2019-09-07
i18n

sjawhar:
## Purpose
Lausanne is in French Switzerland, yo!

## Approach
* Moved all copy for signup into languages files
* Added French translations to everything in Signup
* Added ability to switch back and forth between languages in Signup
* Signup language selection is persisted and effects the rest of authentication components


sjawhar:
I'll add a priority Q3 task to make a new sign in page. Right now we're using the built-in Amplify page, and I really don't feel like making a new page when that one is working :laughing:

PR #77, opened 2019-09-14
add pre-signup lambda trigger

Borchardt:
## Purpose
Enable auto-confirmation of users on signup

## Approach
Add pre-signup Lambda trigger to user pool
Handler code and test input taken from [this AWS example](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-sign-up.html#aws-lambda-triggers-pre-registration-example-2)

## Testing
You can test against my development environment by using the following `.env` settings:

```
AMPLIFY_REGION=us-east-1
AMPLIFY_USER_POOL_ID=us-east-1_k74dQxKWS
AMPLIFY_USER_POOL_WEB_CLIENT_ID=4i423hjgrijienq6mepug3gnrk
CONCERT_START_TIME=3000
WEBSOCKET_API_URL=wss://b7o502lol0.execute-api.us-east-1.amazonaws.com/Prod
WEBSOCKET_API_STUB=false
```

Testing steps:
1: begin sign up with a new phone number
2: after submission of user information and before submitting confirmation, navigate back to sign in screen
3: sign in with unconfirmed phone number

- [ ] Do we also want to auto-verify phone numbers? (Doesn't seem necessary or right, but easy to add)
- [ ] Do any more non-client changes need to be made in order to enable removing confirmation flow from the app?


users/cloudformation.yml
```diff
@@ -56,6 +58,17 @@ Resources:
                 Effect: Allow
                 Resource: '*'
 
+  PreSignUpFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      FunctionName: !Sub ${Environment}-enchanted-brain-pre-signup
+      Runtime: python3.7
+      CodeUri: ../concert/app/src/functions/pre_signup
```
sjawhar: The function should be in the users folder, not concert

users/cloudformation.yml
```diff
@@ -123,6 +138,16 @@ Resources:
             MinLength: 2
             MaxLength: 3
 
+  UserPoolLambdaInvokePermission:
+    Type: AWS::Lambda::Permission
+    Properties:
+      Action: lambda:invokeFunction
+      Principal: cognito-idp.amazonaws.com
+      FunctionName: !Ref PreSignUpFunction.Alias
+      SourceArn: !Sub
```
sjawhar:
```suggestion
      SourceArn: !Sub arn:aws:cognito-idp:${AWS::Region}:${AWS::AccountId}:userpool/${CognitoUserPool}
```

users/cloudformation.yml
```diff
@@ -142,7 +142,7 @@ Resources:
       Action: lambda:invokeFunction
       Principal: cognito-idp.amazonaws.com
       FunctionName: !Ref PreSignUpFunction.Alias
-      SourceArn: !Sub arn:aws:cognito-idp:${AWS::Region}:${AWS::AccountId}:userpool/${CognitoUserPool}
+      SourceArn: !GetAtt CognitoUserPool.Arn
```
Borchardt: :mind_blown:

PR #95, opened 2020-01-19
Feature/86 play song

### LINKED ISSUE #86: The user listens to a song and picks emotions
## Expected Behavior
* After filling out demographic information, the user should see instructions for the emotion picker. There should be a button to begin.
* When the begin button is pressed, [song playback begins](https://docs.expo.io/versions/latest/sdk/audio/) and the user is prompted for emotions in the same way as concert attendees were.
* New environment variables should control the song and emotion the user is given:
  * `MTURK_SONG_ID`: string corresponding to one of the four songs in [this table in the wiki](https://github.com/sjawhar/enchanted-brain/wiki/Requirements#part-2-coloremotion-1)
  * `MTURK_CHOICE_TYPE`: "CHOICE_EMOTION_HAPPINESS" or "CHOICE_EMOTION_ANGER"
  * `MTURK_CHOICE_INTERVAL`: 20
  * `MTURK_CHOICE_TIMEOUT`: 5
  * The `choiceInverted` option should be set randomly (50/50)
* Song choices should be persisted to the state, but there is currently a mechanism which also sends the choice to concerApi. This should not be called in the MTurk app.

asyrtsova:
## Purpose
Closes #86


client/app/src/state/actions.js
```diff
@@ -1,10 +1,12 @@
 import concertApi from '../api/concertApi';
 import { CHOICE_CHILLS } from '../constants/Choices';
 import { CHOICE_MADE } from '../constants/Events';
+import { AMPLIFY_AUTH_DISABLE } from 'react-native-dotenv'; //to check if MTURK version
```
sjawhar:
```suggestion
import { AMPLIFY_AUTH_DISABLE } from 'react-native-dotenv';
```
This is not a useful comment, it's duplicated below

client/app/src/constants/Messages.js
```diff
@@ -1,6 +1,6 @@
-export const MESSAGE_INSTRUCTION_COLOR =
+export const MESSAGE_PROMPT_COLOR =
```
sjawhar: Please keep all these constants organized alphabetically

PR #96, opened 2020-02-05
Feature/88 post data

### LINKED ISSUE #88: User choices are submitted and the user is thanked
## Expected Behavior
* When the song the user is listening to (#86) is finished, user and choice data should be POSTed to the API (#87).
* When the data is sent, the user should be shown a thank you screen with a unique code (e.g. uuidv4). The user will then provide this code to MTurk to validate their submission, and we can use the code to identify legitimate respondants.

asyrtsova:
## Purpose
Closes #88 

## Screenshots
Thank You Screen:
Screenshot_20200205-104736


client/app/src/screens/ThankYou.js
```diff
@@ -0,0 +1,82 @@
...
+        demographics,
+        choices,
+      }),
+    };
+
+    if (MTURK_API_STUB === 'true') {
+      console.log(fetchParams);
+      return;
+    }
+    fetch(MTURK_API_URL, fetchParams);
```
sjawhar: Errors here need to be handled, otherwise the app will crash.

client/app/src/screens/Synesthesia.js
```diff
@@ -60,15 +59,7 @@ export default class SynesthsiaScreen extends Component {
       });
       return;
     }
-    InteractionManager.runAfterInteractions(() =>
```
sjawhar: For the concert app to stay operational, this deleted functionality now needs to be passed in as the onEnd() function here: https://github.com/sjawhar/enchanted-brain/blob/c4bab360d6178de6b03c29ff78a82054bfe44ec0/client/app/src/App.js#L176-L184

client/app/src/constants/Messages.js
```diff
@@ -18,6 +18,9 @@ export const MESSAGE_RESPONSE_RECORDED_HEADER = 'Response recorded';
 export const MESSAGE_STAGE_COMPLETE_BODY = 'Please enjoy the next stage of the concert.';
 export const MESSAGE_STAGE_COMPLETE_HEADER = 'Get Enchanted!';
 
+export const MESSAGE_THANK_YOU_HEADER = 'Thank You!';
+export const MESSAGE_THANK_YOU_BODY = 'Copy the following ID into MTurk';
```
sjawhar:
```suggestion
export const MESSAGE_THANK_YOU_BODY = 'Your responses have been recorded. Enter the code below into MTurk to complete the assignment.;
```

client/app/src/screens/Synesthesia.js
asyrtsova: I changed "this.props.navigation.navigate" to "NavigationService.navigate" to use that code in App.js. Let me know if that's incorrect/ will cause issues.

client/app/src/App.js
```diff
@@ -181,6 +182,15 @@ class App extends React.Component {
         onDisconnect: canConnect && this.handleDisconnect,
         stageId,
         ...stageData,
+        onEnd: InteractionManager.runAfterInteractions(() =>
```
sjawhar: This needs to be wrapped in a callback:
```suggestion
        onEnd: () => InteractionManager.runAfterInteractions(() =>
```

client/app/src/screens/ThankYou.js
```diff
@@ -0,0 +1,91 @@
...
+
+    fetch(MTURK_API_URL, fetchParams)
+      .then(response => {
+        if (!response.ok) {
+          console.debug('FETCH response status:', response.status);
+        }
+      })
+      .catch(error => {
+        console.debug('FETCH error:', error);
+      });
```
sjawhar: It would be more consistent with the rest of our app if this were done using async/await and try/catch:
```js
try {
  const response = await fetch(...);
  if (!response.ok) {
    throw response;
  }
} catch (error) {
  // Do stuff
}
```

